PR #35, opened 2018-11-15
Moving the location and department fields and vocabularies to tide_core

kurtfoster:


config/install/taxonomy.vocabulary.department.yml
```diff
@@ -0,0 +1,8 @@
+langcode: en
+status: true
+dependencies: {  }
+name: Department
+vid: department
+description: 'Department for media type.'
```
GROwen: @kurtfoster Is this description still accurate?

PR #39, opened 2018-11-26
SDPA-309] Update save config functions

enzolutions:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-309

Content SSP PR: https://github.com/dpc-sdp/content-ssp-vic-gov-au/pull/44

Content PR: https://github.com/dpc-sdp/content-vic-gov-au/pull/346

Tide Media PR: https://github.com/dpc-sdp/tide_media/pull/6

### Changed

1.  Update helper to save files in the disk
2. Add function to delete config and delete from disk if is necessary

### Screenshots


includes/helpers.inc
```diff
@@ -47,6 +58,63 @@ function _tide_read_config($id, array $locations = []) {
...
+ * @param bool $file_system_delete
+ *   Determine if config file should be deleted from config sync directory.
+ *
+ * @return bool
+ *   Indicates success or failure of the function.
+ *
+ * @throws \Exception
+ *   If configuration file was not deleted successfully.
+ */
+function _tide_delete_config($id, array $locations = [], $file_system_delete = FALSE) {
```
sonnykt: If you remove a file from `config/sync` in your commit, it will be removed upon deployment, and the config will be removed during config-import. Deleting config files via code won't work.

anthony-malkoun: @sonnykt This did not get actioned. Is this change you've requested something you can do so we can get this merged?

PR #63, opened 2019-02-26
[VSM-187] Added drush command to import bulk redirects

FleetAdmiralButter:


tide_core.drush.inc
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Implements hook_drush_command().
+ */
+function tide_core_drush_command() {
+  $items = [];
+
+  $items['redirect-import'] = [
+    'description' => "Imports a file and adds redirects to the redirect table",
+    'arguments' => [
+      'filepath' => 'Path to CSV (comma-separated), JSON or PHP file containing old and new URLs for redirects',
```
anthony-malkoun: Can this be a URI?

tide_core.drush.inc
```diff
@@ -0,0 +1,188 @@
...
+  ];
+
+  return $items;
+
+};
+
+/**
+ * Get the node id of the page to redirect to.
+ */
+function get_node_id($to) {
```
anthony-malkoun: is this function getting a node id? It seems to be normalising URLs if anything if I understand it correctly.

tide_core.drush.inc
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Uses redirect entity to save entry to DB.
+ */
+function add_redirects($from, $to) {
+  if ($from != '' && $to != '') {
+    $redirectrepo = redirect_repository();
+    if ($redirectrepo->findBySourcePath($from)) {
+      return;
+    }
+    $vicgov = $to == 'https://vic.gov.au' || $to == 'http://vic.gov.au' || $to == 'vic.gov.au';
```
anthony-malkoun: This check is for just the homepage URLs? What happens if they come through with a trailing slash?

tide_core.drush.inc
```diff
@@ -0,0 +1,188 @@
...
+    $wwwvicgov = $to == 'https://www.vic.gov.au' || $to == 'http://www.vic.gov.au' || $to == 'www.vic.gov.au';
+    if ($vicgov || $wwwvicgov) {
+      $to = 'site-4/';
```
anthony-malkoun: So `$to` is, I guess, a path prefix for certain URLs? Maybe that should just be a lookup in an array of redirect exceptions rather than the conditional statements above?

anthony-malkoun: Ah no, looking at this, it is rewriting the whole `$to` value :+1:

tide_core.drush.inc
```diff
@@ -0,0 +1,188 @@
...
+    $queryStr = substr($from, 0, strpos($from, '?'));
+    if (!empty($queryStr)) {
+      $query = explode('&', $queryStr);
+    }
+    else {
+      $query = [];
+    }
+    $redirect->setSource($from, $query);
+    $redirect->setRedirect(get_node_id($to));
+    $redirect->getRedirect();
```
anthony-malkoun: Why is this required?

anthony-malkoun:
Hey @FleetAdmiralButter I was thinking about how to do this... would a migration be a good fit for this? Include `migration_csv` and I *think* redirects are supported as a destinations in a migration. Then could use drush to maintain them. This may be a completely bonkers idea :)

FleetAdmiralButter:
> Just some random thoughts... These aren't things you need to change but more me getting an idea of how things are done around here :)
> 
> * Can we run `lint` locally? Just trying to get my head around why it took multiple commits to get it fixed so I suspect that we can't?

We should be able to run lint locally using `ahoy lint` but I haven't been able to get mine to run 😞, so that's why I ended up pushing the commits.

> * How do I test this? It would be good if some testing steps went into the description.

QA Steps would usually be in the description of the linked ticket on JIRA.

tide_core.drush.inc
FleetAdmiralButter: Supports both, full URL is used for redirecting from or to external sites but if either is internal the entry would just specify the internal URI link.

tide_core.drush.inc
```diff
@@ -21,7 +35,302 @@ function tide_core_drush_command() {
...
+ *   Status code to return for redirect. Overridden if $to specifies code.
+ *
+ * @return bool
+ *   returns FALSE if input was invalid or a node id could not be found.
+ *
+ * @throws \Exception
+ *   If saving to the DB fails.
+ */
+function add_redirects($from, $to, $home_url, $lang, $status_code) {
+
```
alexdesignworks: @anthony-malkoun how did this pass phpcs? Could you please check that this file is in a list of lint targets?
There should be no new line after `function`

anthony-malkoun:
@FleetAdmiralButter what's happening with this PR?

tide_core.drush.inc
FleetAdmiralButter: This has been fixed @AlexSkrypnyk

PR #57, opened 2019-02-01
[SDPA-1213] Upgraded Metatag.

sonnykt:
JIRA: https://digital-engagement.atlassian.net/browse/SDPA-1213

Related PRs:
* Tide Core: https://github.com/dpc-sdp/tide_core/pull/57
* Tide Media: https://github.com/dpc-sdp/tide_media/pull/19
* Tide API: https://github.com/dpc-sdp/tide_api/pull/28
* Tide Site: https://github.com/dpc-sdp/tide_site/pull/18
* Tide Webform: https://github.com/dpc-sdp/tide_webform/pull/21
* Tide News: https://github.com/dpc-sdp/tide_news/pull/14
* Tide Page: https://github.com/dpc-sdp/tide_page/pull/14
* Tide Event: https://github.com/dpc-sdp/tide_event/pull/24
* Tide Landing Page: https://github.com/dpc-sdp/tide_landing_page/pull/63
* Tide Monsido: https://github.com/dpc-sdp/tide_monsido/pull/13
* Tide Search: https://github.com/dpc-sdp/tide_search/pull/11
* Tide: https://github.com/dpc-sdp/tide/pull/39
* Content-VIC: https://github.com/dpc-sdp/content-vic-gov-au/pull/462

PR #66, opened 2019-03-05
[SDPA-1207] Added content archive operation.

barbun:
Story: https://digital-engagement.atlassian.net/browse/SDPA-1207

### Changed:
- Pinned Metatag module version to stop the patch breaking the build (known issue, see #57 )
- Added new custom confirmation Form for entity Archive operation.
- Added custom route that points to the confirmation form mentioned above.
- Added hook_entity_operation that adds the dropdown option "Archive"

### Note
- I was unsuccessful with adding Behat tests due to permissions that Workflow generates dynamically. The module checks for "use editorial transition archived" which does not seem to exist during Behat testing.

### Screenshots
screen shot 2019-03-06 at 10 46 33 am

### DIFF
--- a/src/Form/EntityArchiveForm.php
+++ b/src/Form/EntityArchiveForm.php
@@ -1,0 +1,141 @@
+<?php
+
+namespace Drupal\tide_core\Form;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Form\ConfirmFormBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Messenger\MessengerInterface;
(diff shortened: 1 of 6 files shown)


src/Form/EntityArchiveForm.php
```diff
@@ -0,0 +1,141 @@
...
+   */
+  public function getFormId() {
+    return 'tide_core_entity_archive';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getQuestion() {
+    return $this->t('Are you sure you want to archive entity %name?', ['%name' => $this->entity->label()]);
```
sonnykt: The confirmation message is not necessary but can stay at this stage as we lack of resource to fix it.

PR #68, opened 2019-05-13
Reworked config import utility.

sonnykt:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -103,7 +103,7 @@
                 "Scheduled Updates does not work with Content Moderation - https://www.drupal.org/project/scheduled_updates/issues/2821916": "https://www.drupal.org/files/issues/scheduled_updates_does-2821916-7.patch"
             },
             "drupal/field_group": {
-                "Do not translate field group labels using interface translation. - https://www.drupal.org/project/field_group/issues/2846617": "https://www.drupal.org/files/issues/2018-05-10/2846617-16-use-config-translations-for-field-group.patch"
+                "Do not translate field group labels using interface translation. - https://www.drupal.org/project/field_group/issues/2846617": "https://www.drupal.org/files/issues/2019-04-18/field_group-use_config_translation-2846617-26.patch"
             }
         }
     }
--- a/dev-tools.sh
+++ b/dev-tools.sh
@@ -33,6 +33,6 @@
 # stable version.
 #
 # Uncomment and set the Dev-Tools's commit value and commit this change.
-# export GH_COMMIT=COMMIT_SHA
+export GH_COMMIT=b978f632476ea8c6c7d4e69c149fe637297dbf88
 
 bash <(curl -L https://raw.githubusercontent.com/dpc-sdp/dev-tools/master/install?"$(date +%s)") "$@"
--- a/dev-tools.sh
+++ b/dev-tools.sh
@@ -33,6 +33,6 @@
 # stable version.
 #
 # Uncomment and set the Dev-Tools's commit value and commit this change.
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
         "drupal/config_split": "^1.4",
         "drupal/config_update": "^1.6",
         "drupal/content_moderation_scheduled_updates": "^1.0",
-        "drupal/core": "8.6.12",
+        "drupal/core": "8.6.16",
```
anthony-malkoun: @GROwen could you double check that you merged hotfix updates back to develop as this should not be required.

includes/helpers.inc
```diff
@@ -58,66 +57,6 @@ function _tide_read_config($id, array $locations = [], $override = FALSE) {
   return $storages[$path]->read($id);
 }
 
-/**
```
anthony-malkoun: I assume this wasn't being used and is safe to remove @sonnykt. Is there another reason it is being removed? Would it actually work as deleting the file is probably not enough as the next config export will reinstate the files?

includes/helpers.inc
```diff
@@ -206,16 +143,6 @@ function _tide_import_single_config($config_name, array $locations = [], $overri
   try {
     $config_importer->import();
     \Drupal::cache('config')->delete($config_name);
-    // If configuration was set to override, then ensure it is written to active
-    // config.
-    if ($override) {
```
anthony-malkoun: Is this removed for the same reason that the delete function probably doesn't work?

includes/helpers.inc
sonnykt: @anthony-malkoun it was added in the previous implementation of SDPA-309, which was not correct. It only works on env having write permission to config files - aka local only. It has never worked on lagoon as the config files are managed by git deployment.

includes/helpers.inc
sonnykt: same, removed what were added by the previous implementation of SDPA-309.

alexdesignworks: I wanted to point out that this code caused a problem with deployments. This is due to the fact that it was re-generating the same existing config with new UUIDs in different environment. So, the config was correctly exported in local, passed ci and develop env, but would not deploy to production since UUIDs would be re-generated as a part of the update, which runs _before_ config import.

Not sure what you are doing with this code right now - moving it to another place etc. - but there should be a flag to allow to skip writing of the config.

includes/helpers.inc
anthony-malkoun: That makes sense to me @sonnykt. Thanks for explaining.

includes/helpers.inc
anthony-malkoun: If I understand it correctly, it is being removed completely. Writing of the config should never be done as it will never work outside of a local environment @alexdesignworks.

dev-tools.sh
```diff
@@ -33,6 +33,6 @@
 # stable version.
 #
 # Uncomment and set the Dev-Tools's commit value and commit this change.
-# export GH_COMMIT=COMMIT_SHA
+export GH_COMMIT=b978f632476ea8c6c7d4e69c149fe637297dbf88
```
anthony-malkoun: This will be removed when dev tools is merged back.

composer.json
GROwen: I'm not sure why this is showing as a diff @anthony-malkoun. The update to 8.6.16 is in both master and develop branches.

GROwen: @anthony-malkoun running a rebase has sorted it out.

PR #76, opened 2019-07-02
Add PoC for composer plugin to install platform tools

anthony-malkoun:
This is a very quick PoC to determine whether using a composer plugin style process would work to allow a single workflow for local builds and for starter kit/scaffold builds using a well known paradigm of `composer require dpc-sdp/sdp-platform`.

Naming is intentionally removed from `tide` as this may also be used for hosting only applications.

This is draft/WIP/Poc only. If this method is appropriate it will be moved out to its own repo.


composer.json
```diff
@@ -6,58 +6,16 @@
...
-        "drupal/simple_sitemap": "^2.12",
-        "drupal/smart_trim": "^1.1",
-        "drupal/stage_file_proxy": "^1.0-alpha3",
-        "drupal/token_conditions": "dev-1.x",
-        "drupal/token_filter": "^1.1",
-        "drupal/twig_field_value": "^1.1",
-        "drupal/views_menu_children_filter": "^1.0-beta2",
-        "drupal/viewsreference": "^2.0-alpha4",
-        "drupal/wysiwyg_template": "^2.0@beta"
+        "dpc-sdp/sdp-platform": "^0.0.1"
```
anthony-malkoun: Ignore changes here as I've removed everything to test just the plugin.

dpc-sdp/sdp-platform/assets/.docker/Dockerfile.cli
```diff
@@ -0,0 +1,33 @@
+# @see https://github.com/dpc-sdp/bay/blob/master/bay/images/Dockerfile.php
+FROM singledigital/bay-cli:latest
+ARG COMPOSER
+
+ENV WEBROOT=docroot \
+    COMPOSER_ALLOW_SUPERUSER=1 \
+    COMPOSER_CACHE_DIR=/tmp/.composer/cache \
+    MYSQL_HOST=mariadb \
+    COMPOSER=${COMPOSER:-composer.json} \
+    DRUPAL_MODULE_PREFIX=%%DRUPAL_MODULE_PREFIX%%
```
anthony-malkoun: The string `%%DRUPAL_MODULE_PREFIX%%` will be replaced during install/update. Ultimately this should be from something like `.sdp-config.yml` or similar but is hard wired now just to test it works.

dpc-sdp/sdp-platform/src/composer/Plugin.php
```diff
@@ -0,0 +1,79 @@
...
+      ScriptEvents::POST_UPDATE_CMD => 'addPlatformFiles',
+    );
+  }
+
+  /**
+   * Add/update platform files.
+   *
+   * @param \Composer\Script\Event $event
+   */
+  public function addPlatformFiles(Event $event) {
```
anthony-malkoun: This is obviously very brute force for the minute... it might make sense to still use ansible for some of this. Thoughts on how possible/easy to follow this would be?

PR #79, opened 2019-07-15
[SDPA-2742] remove scheduled_updates module

vincent-gao:
https://digital-engagement.atlassian.net/browse/SDPA-2742


vincent-gao:
@anthony-malkoun The `build_suggest` tests passed on my local build.  
image

image

and it also passed my manual test.
image

vincent-gao:
ohhhhh, it passes...

composer.json
```diff
@@ -42,7 +41,6 @@
         "drupal/prlp": "^1.3",
         "drupal/redirect": "^1.3",
         "drupal/role_delegation": "^1.0-alpha1",
-        "drupal/scheduled_updates": "^1.0-alpha7",
```
anthony-malkoun: @vincent-gao do we need an update hook to disable the module? Have we tried this on a clean install and checked that it doesn't complain about module missing from file system?

tide_core.info.yml
```diff
@@ -70,6 +68,8 @@ dependencies:
   - viewsreference:viewsreference
   - wysiwyg_template:wysiwyg_template
   - wysiwyg_template:wysiwyg_template_core
+  - drupal:scheduled_transitions
```
anthony-malkoun: Why do these start with `drupal` and aren't of the form `scheduled_transitions:scheduled_transitions`?

tide_core.install
```diff
@@ -225,3 +225,17 @@ function tide_core_update_8006() {
     }
   }
 }
+
```
anthony-malkoun: @vincent-gao do we need update hooks to disable the module (Will it just remove the tables in this case, so the update below not needed) and a future ticket to remove from `composer.json`? Also, should we be using the field API to remove the fields that are marked as removed in the yml files?

composer.json
sonnykt: yes we need an update hook to uninstall the module. Make sure that the module is removed from all dependencies.

tide_core.info.yml
vincent-gao: changed.

composer.json
vincent-gao: yes. updated an update_hook for this purpose

tide_core.install
vincent-gao: The schduled_updats configs are spread across tide modules. and all of them refer to the tide_test and tide_core module. this is why I dealt with `tide_test` and `tide_core` first.

here is the plan:
1. the update hook for deleting and uninstalling scheduled_updates module will be in tide_core
2. delete all of the scheduled_updates configs from other tide modules like tide_page, tide_landing_page ...

We don't need to use field API to remove the fields I think. after `schduled_updats` module gets removed from all dependencies, we will have an installation without schduled_updats.

vincent-gao: the test failed, but I think this is c.

sonnykt: @vincent-gao Scheduled Updates fields of all content types depend on the `scheduled_updates.scheduled_update_type.*` Scheduled Update Types. Those Types are entities defined by the Scheduled Updates module.

From my understanding, you cannot uninstall a module if there are existing entities defined by that module.

vincent-gao: @sonnykt yes. that is right. we have an update_hook for deleting all existing entities(only 2 scheduled entities).

vincent-gao: @sonnykt after this gets merged to develop, I will delete dependencies and configs from other tide submodules. finally dealing with client sites like content-vic, actually content-vic is already modified by https://github.com/dpc-sdp/content-vic-gov-au/pull/616

PR #97, opened 2019-10-06
[SDPA-3230] Fixed content save fails issue

vincent-gao:
Content save fails on CMS for Site Admin and below if Related links contains a link to a deleted page
jira: https://digital-engagement.atlassian.net/projects/SDPA/issues/SDPA-3230
### Issue
1. On a new or existing Landing Page, go to the Related Links section and add a link to an internal page.

2. Save.

3. Delete the page that was linked to in the Related Links section.

4. Using a Site Admin account or below, edit the previously created Landing Page.

5. When saving the page, a "The path 'entity:node/2510' is inaccessible" error occurs.
image

### expected result
The save should succeed as it does with an Administrator account, or the error message should be more descriptive and point the user to the broken Related Link entry. Update with the following error message.

'A link on this page is broken. Please update or remove the link and the page will save.'

### actual result
Save fails with a generic error.

related PR : https://github.com/dpc-sdp/content-vic-gov-au/pull/719

### DIFF
--- a/tide_core.info.yml
+++ b/tide_core.info.yml
@@ -144,8 +144,6 @@ config_devel:
     - metatag.metatag_defaults.node
     - metatag.metatag_defaults.taxonomy_term
     - metatag.metatag_defaults.user
-    - paragraphs.paragraphs_type.timeline
-    - paragraphs.paragraphs_type.timelines
     - pathauto.pattern.content_title
     - pathauto.settings
     - prlp.settings


tide_core.module
```diff
@@ -189,3 +191,45 @@ function tide_core_entity_bundle_create($entity_type_id, $bundle) {
...
+/**
+ * Used by hook_ENTITY_TYPE_delete.
+ *
+ * When deleting a node randomly, we need to update the referencing node field
+ * and delete the Paragraph entity that the deleted node attached to.
+ */
+function _tide_core_clean_up_related_links($entity) {
+  // Instead of using Paragraph getParent function, we use
+  // database api directly, which is easy to read and understand.
+  $query = \Drupal::database()->select('node__field_related_links', 'n');
```
sonnykt: @vincent-gao cc @anthony-malkoun @GROwen 
I'm extremely concerned about the performance impact of this solution. Upon any entity deletion, we join 2 large tables to find if anything references the to-be-deleted entity in *its link field*, then update them accordingly. Drupal is not supposed to behave that way.

IMO the error message from Drupal is correct as expected due to the link fields are configured to validate the destination links.

GROwen: I agree with @sonnykt - another solution could be to alter the default system message with something a little more helpful.

vincent-gao: Yes~~~. I agree. The error message is definitely correct and expected.
The validation logic is hidden under function`massageFormValues` from `contrib/paragraphs/src/Plugin/Field/FieldWidget/ParagraphsWidget.php`  `line 2274`.
there is no opportunity to alter this error message. or replaced the default widget(insane and no need).

Or

could be fixing by deleting the attached paragraph entity by using the entity query.

PR #101, opened 2019-10-20
[SDPA-3230] alter error message

vincent-gao:
Content save fails on CMS for Site Admin and below if Related links contains a link to a deleted page
jira: https://digital-engagement.atlassian.net/projects/SDPA/issues/SDPA-3230

Issue
On a new or existing Landing Page, go to the Related Links section and add a link to an internal page.

Save.

Delete the page that was linked to in the Related Links section.

Using a Site Admin account or below, edit the previously created Landing Page.

When saving the page, a "The path 'entity:node/2510' is inaccessible" error occurs.
image
image

expected result
The save should succeed as it does with an Administrator account, or the error message should be more descriptive and point the user to the broken Related Link entry. Update with the following error message.

'A link on this page is broken. Please update or remove the link and the page will save.'

actual result
Save fails with a generic error.

related PR : dpc-sdp/content-vic-gov-au#719


tide_core.module
```diff
@@ -189,3 +190,19 @@ function tide_core_entity_bundle_create($entity_type_id, $bundle) {
...
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function tide_core_preprocess_status_messages(&$variables) {
+  if (isset($variables['message_list']['error']) && !empty($variables['message_list']['error'])) {
+    foreach ($variables['message_list']['error'] as &$error_message) {
+      if ($error_message instanceof Markup) {
+        $error_message = (string) $error_message->__toString();
+        if (strpos($error_message, 'Validation error on collapsed paragraph field_paragraph_link') !== FALSE) {
+          $error_message = t('A link in <i>Related links</i> field on this page is broken. Please update or remove the link and the page will save.');
```
GROwen:
```suggestion
          $error_message = t('A link in <i>Related links</i> field on this page is broken. Please update or remove the link and retry.');
```

I'm suggesting this change because we can't guarantee that the node will save once the user has made a change to the Related Links field.

tide_core.module
```diff
@@ -189,3 +190,19 @@ function tide_core_entity_bundle_create($entity_type_id, $bundle) {
     }
   }
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function tide_core_preprocess_status_messages(&$variables) {
+  if (isset($variables['message_list']['error']) && !empty($variables['message_list']['error'])) {
```
GROwen: We probably want to get more specific with the context of when this message is altered i.e. only when it relates to `Related links`

vincent-gao: Hi @GROwen , I added more specific checkers, which will make sure it only happens under nod context.

tide_core.module
vincent-gao: changed and committed.

tide_core.module
vincent-gao: @GROwen I think the new condition that I added will be enough for checking the context.

GROwen:
@vincent-gao code looks good but CI needs fixing.

vincent-gao:
> @vincent-gao code looks good but CI needs fixing.

@GROwen yes it is. ~ sad.

vincent-gao:
> @vincent-gao code looks good but CI needs fixing.

@GROwen , It is still getting failed by `Unable to complete AJAX request. (RuntimeException)`. 
but In my local env all tests are good to go. moreover, in content-vic https://nginx-php-content-vic-pr-719.lagoon.vicsdp.amazee.io/media/add/document , I can get an expected result.

PR #103, opened 2019-10-23
[SDPA-146] Added configuration update hook for seckit module.

MdNadimHossain:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-146

### Changed

1.  Added update hook to update the seckit module settings.

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -372,3 +372,26 @@ function tide_core_update_8014() {
     $module_installer->install(['link_field_autocomplete_filter']);
   }
 }
+
+/**
+ * Config for Seckit module.
+ */
+function tide_core_update_8015() {
+  if (\Drupal::moduleHandler()->moduleExists('seckit')) {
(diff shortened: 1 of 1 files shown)


tide_core.install
```diff
@@ -372,3 +372,26 @@ function tide_core_update_8014() {
     $module_installer->install(['link_field_autocomplete_filter']);
   }
 }
+
+/**
+ * Config for Seckit module.
+ */
+function tide_core_update_8015() {
+  if (\Drupal::moduleHandler()->moduleExists('seckit')) {
```
kurtfoster: @MdNadimHossain , you need to enable it if it's not enabled. It's already in composer.json, so we know that it is in the codebase.

MdNadimHossain: Hi @kurtfoster , As we discussed, it is added as a dependency in this module so I am only checking if the module exists and then updating the config.

PR #105, opened 2019-11-04
[SDPSUP-1082] Added clamav config to use central service.

MdNadimHossain:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPSUP-1082

### Changed

1. Updated clamav.settings in optional
2. Added update hook to change clamav config to use central service.

### DIFF
--- a/config/optional/clamav.settings.yml
+++ b/config/optional/clamav.settings.yml
@@ -1,13 +1,13 @@
 enabled: true
 outage_action: 0
 overridden_schemes: {  }
-scan_mode: 1
+scan_mode: 0
 verbosity: 0
 mode_executable:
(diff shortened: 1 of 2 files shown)


config/optional/clamav.settings.yml
```diff
@@ -1,13 +1,13 @@
 enabled: true
 outage_action: 0
 overridden_schemes: {  }
-scan_mode: 1
+scan_mode: 0
```
anthony-malkoun: Is this the setting that puts it into TCP/IP mode?

MdNadimHossain: Yes

PR #125, opened 2020-02-09
[SDPA-3779] fixes unexpected element

vincent-gao:
## Jira
https://digital-engagement.atlassian.net/browse/SDPA-3779

## Changes
1. the form_alter only happens when the element of['revision_log'] exists.

image

### DIFF
--- a/tide_core.module
+++ b/tide_core.module
@@ -216,9 +216,11 @@ function tide_core_preprocess_status_messages(&$variables) {
  * Implements hook_form_FORM_ID_alter().
  */
 function tide_core_form_content_moderation_entity_moderation_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-  $form['#attached']['library'] = 'tide_core/content_moderation';
-  $form['revision_log']['#type'] = 'textarea';
-  $form['revision_log']['#title'] = t('Publishing instructions and comments');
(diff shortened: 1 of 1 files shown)


tide_core.module
```diff
@@ -216,9 +216,11 @@ function tide_core_preprocess_status_messages(&$variables) {
...
-  $form['revision_log']['#type'] = 'textarea';
-  $form['revision_log']['#title'] = t('Publishing instructions and comments');
-  $form['revision_log']['#description'] = t('All content will be published within 24 hours. State below if your content is embargoed or time critical.');
-  $form['new_state']['#title'] = t('Change status to');
+  if (isset($form['revision_log'])) {
+    $form['#attached']['library'] = 'tide_core/content_moderation';
+    $form['revision_log']['#type'] = 'textarea';
+    $form['revision_log']['#title'] = t('Publishing instructions and comments');
+    $form['revision_log']['#description'] = t('All content will be published within 24 hours. State below if your content is embargoed or time critical.');
+    $form['new_state']['#title'] = t('Change status to');
```
anthony-malkoun, line 224: Should this line be in the `if`?

vincent-gao: yes~~~.  I think so 😧

anthony-malkoun: Where does it appear? I can't see it in the screen shot. It just doesn't feel like it's related to the `revision_log` but if you're sure it should be there then all good.

PR #133, opened 2020-04-29
[SDPA-4029] revokes permission

vincent-gao:
## JIRA
https://digital-engagement.atlassian.net/browse/SDPA-4029

## CHANGES
1. delete `administer taxonomy` from yml files.
2. add `access taxonomy overview` to yml files.
3. an hook_update for revoking  `administer taxonomy`  and granting `access taxonomy overview` from drupal site
4.  approver and site_admin both can still edit or create new terms.

https://github.com/dpc-sdp/content-vic-gov-au/pull/847

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -442,3 +442,17 @@ function tide_core_update_8018() {
   $config->set('display.default.display_options.empty.area.content.value', 'There are no scheduled updates yet.');
   $config->save();
 }
+
+/**
+ * Change permissions.
+ *
+ * Revokes 'administer taxonomy' permission from site_admin and approver roles,
+ * and grants them 'access taxonomy overview'.
+ */
+function tide_core_update_8019() {
+  $roles = ['approver', 'site_admin'];
+  foreach ($roles as $role_id) {
+    user_role_revoke_permissions($role_id, ['administer taxonomy']);
+    user_role_grant_permissions($role_id, ['access taxonomy overview']);
+  }
+}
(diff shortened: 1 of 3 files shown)


anthony-malkoun:
> @vincent-gao My reading of the issue is that delete access should be removed from the Sites vocab only. This doesn't look like it's targeting the sites vocab specifically but all vocabs.

I think that's right @GROwen as then the normal permissions will take effect. The permission being granted is the equivalent of `view content` for taxonomy terms.

tide_core.install
```diff
@@ -442,3 +443,25 @@ function tide_core_update_8018() {
...
+ *
+ * Revokes 'administer taxonomy' permission from site_admin and approver roles,
+ * and grants them 'access taxonomy overview'.
+ */
+function tide_core_update_8019() {
+  $roles = ['approver', 'site_admin'];
+  $taxonomy_permissions = new TaxonomyPermissions(\Drupal::entityManager());
+  foreach ($roles as $role_id) {
+    $assigned_permissions = ['access taxonomy overview'];
+    foreach ($taxonomy_permissions->permissions() as $permission => $details) {
```
anthony-malkoun, line 458: What are you doing here? Are you removing the `delete` permission?

anthony-malkoun:
> > @vincent-gao My reading of the issue is that delete access should be removed from the Sites vocab only. This doesn't look like it's targeting the sites vocab specifically but all vocabs.
> 
> I think that's right @GROwen as then the normal permissions will take effect. The permission being granted is the equivalent of `view content` for taxonomy terms.

Actually I think @GROwen is right (I seem to be making a habit of this)... I don't think you should be trying to modify existing permissions beyond doing what you need to do.

GROwen:
@vincent-gao With the opening of dpc-sdp/tide_site#53 can this PR be closed?

vincent-gao:
yes. @GROwen . this one should be closed.

PR #140, opened 2020-05-14
Move install hook to sub module

anthony-malkoun:
Created to allow patching for upgrade of RCMPI


anthony-malkoun:
No longer required

PR #141, opened 2020-05-28
[SDPA-4186]  restricting jwt editing route.

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-4186
### Changes
1. the permissions cannot be changed directly from the permission matrix, as jwt doesn’t have its own permission, it uses system permissions named `Use the administration pages and help`. so we will change it from RouterSubscribe.


vincent-gao:
closed it as we are going to update webform module

PR #144, opened 2020-06-09
SDPA-4313 enable schedule transition for each content type

MdNadimHossain:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-4313
### Changes
1. Added update hook to enable schedule transitions for each content type.
2. Added update hook to remove default password policy, if exists, and replace it with the SDPA password policy.
3. Added patches for the password policy module.

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -593,9 +593,7 @@ function tide_core_update_8021() {
  */
 function tide_core_update_8022() {
   // Get the available content types.
-  $types = \Drupal::entityTypeManager()
-  ->getStorage('node_type')
-  ->loadMultiple();
+  $types = \Drupal::entityTypeManager()->getStorage('node_type')->loadMultiple();
   foreach ($types as $type) {
     if ($type->id()) {
       $content_types[] = $type->id();
--- a/tide_core.install
+++ b/tide_core.install
@@ -587,3 +587,49 @@ function tide_core_update_8021() {
   $config_storage = \Drupal::service('config.storage');
   $config_storage->write($view, $source->read($view));
 }
+
+/**
+ * Enable scheduled transitions for each content type.
+ */
+function tide_core_update_8022() {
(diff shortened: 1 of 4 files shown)


tide_core.install
```diff
@@ -587,3 +587,49 @@ function tide_core_update_8021() {
...
+      $config->set('bundles', $bundles_array);
+      $config->save();
+    }
+  }
+}
+
+/**
+ * Import sdpa password policy.
+ */
+function tide_core_update_8023() {
```
GROwen, line 619: This belongs in dpc-sdp/baywatch because it's SDP specific i.e as an open source project implementations of tide_core may have very different requirements to SDP.

tide_core.install
```diff
@@ -587,3 +587,49 @@ function tide_core_update_8021() {
   $config_storage = \Drupal::service('config.storage');
   $config_storage->write($view, $source->read($view));
 }
+
+/**
+ * Enable scheduled transitions for each content type.
+ */
+function tide_core_update_8022() {
```
GROwen, line 594: I think this logic is better placed in the modules that declare the content types. Alternatively, so that we're not being too prescriptive, it may be better placed in each project.

MdNadimHossain: Cool, I am going to move this hook to each content-type module.

tide_core.install
MdNadimHossain: I am going to move this hook to baywatch module.

tide_core.install
anthony-malkoun: I can be convinced either way. Scheduled transitions should be across the board is my first thought. Will it feel inconsistent if some content types offer it and others don't?

MdNadimHossain: @anthony-malkoun isn't that all the content-type should offer this if schedule transition is enabled?

PR #168, opened 2020-08-20
[SDPA-2970] change display of dates in CMS to Australian style DD/MM/YYYY

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-2970

### Changes
1. apply `d/m/Y - H:i` to the view.

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -783,3 +783,15 @@ function tide_core_update_8030() {
     }
   }
 }
+
+/**
+ * Changes display of dates in CMS to Australian style DD/MM/YYYY.
+ */
+function tide_core_update_8031() {
+  // Changes media view.
+  $media_view = \Drupal::configFactory()->getEditable('views.view.media');
+  $display = $media_view->get('display');
+  $display['default']['display_options']['fields']['changed']['settings']['date_format'] = 'custom';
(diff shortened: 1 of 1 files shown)


GROwen (changes requested):
@vincent-gao This change should happen new installs which means this should probably be added to `tide_core_install()` as well.

vincent-gao:
> @vincent-gao This change should happen new installs which means this should probably be added to `tide_core_install()` as well.

Thanks @GROwen , yes. sure.it should be added to `tide_core_install()`

vincent-gao:
> @vincent-gao This change should happen new installs which means this should probably be added to `tide_core_install()` as well.

Hey @GROwen , I think tide_core(https://github.com/dpc-sdp/tide_core/blob/develop/tide_core.install#L61) has done this for us. during the tide_core module installation, the install hook calls each of `tide_core_update_x` to run.

GROwen (approved):
>Hey @GROwen , I think tide_core(https://github.com/dpc-sdp/tide_core/blob/develop/tide_core.install#L61) has done this for us. during the tide_core module installation, the install hook calls each of tide_core_update_x to run.

Ah yes, we've got an open issue to refactor that logic out. It's not a good pattern to use. This can be cleaned up when that issue is actioned. Thanks @vincent-gao

PR #172, opened 2020-09-03
Added submodule Tide OAuth.

sonnykt:


anthony-malkoun:
@sonnykt for my understanding why would we ever want to provide env variables for generating keys? Is it good enough to let them be generated every time? The main motivation here is that the `AUTHENTICATED_CONTENT` key has been painful to maintain (Lagoon does not handle multiline config variables correctly) and meant that it had to be added manually for PR environments (and remember to add it for new environments). If there is value to having a known value for the keys, can it be something else like the token that TFA uses or does it have to be a full RSA key?

modules/tide_oauth/README.md
```diff
@@ -0,0 +1,50 @@
...
+    $ openssl genrsa -out /tmp/private.key 4096
+    $ openssl rsa -in /tmp/private.key -pubout > /tmp/public.key
+    $ export TIDE_OAUTH_PRIVATE_KEY=`cat /tmp/private.key`
+    $ export TIDE_OAUTH_PUBLIC_KEY=`cat /tmp/public.key`
+    ```
+2. Enable the `tide_oauth` module.
+    * If the TIDE_OAUTH_ environment variables are set, the module will copy
+    the keys to `private://oauth.key` and `private://oauth.pub`.
+    * Otherwise, the module will generate a new key pair.
+3. _(Optional - **Lagoon only**)_ Add a `post-rollout` task to generate the OAuth
```
anthony-malkoun, line 19: Is there any reason this couldn't be an openssl command to generate on the fly from scratch?

modules/tide_oauth/composer.json
```diff
@@ -0,0 +1,12 @@
+{
```
anthony-malkoun, line 1: We are still on Drush 8, does this present an issue here?

sonnykt: it's not an issue. `drush.services.yml` only takes effect from Drush 9. That snippet ensures that we support both 8 and 9.

modules/tide_oauth/README.md
sonnykt: if the key changes, all existing tokens become invalid, even if they are not expired.

sonnykt:
> @sonnykt for my understanding why would we ever want to provide env variables for generating keys? Is it good enough to let them be generated every time? The main motivation here is that the `AUTHENTICATED_CONTENT` key has been painful to maintain (Lagoon does not handle multiline config variables correctly) and meant that it had to be added manually for PR environments (and remember to add it for new environments). If there is value to having a known value for the keys, can it be something else like the token that TFA uses or does it have to be a full RSA key?

If there is no env variable, the module will generate a random key pair upon its installation. The env var is to ensure that all envs have the same key pair.

modules/tide_oauth/README.md
anthony-malkoun: So if the deploy step checked if the key existed first before trying to create it, then it would work?

modules/tide_oauth/composer.json
anthony-malkoun: 👍

modules/tide_oauth/README.md
sonnykt: @anthony-malkoun yes https://github.com/dpc-sdp/tide_core/pull/172/files#diff-6f2392828bd5ca4ff53608718b20927df9c8d6d10a19951924aa36803885473dR51

sonnykt: it only happens upon installation. If it's already on Prod and you launch a new dev env, the key pair doesn't exist on the new env and it still fails. There is no mechanism for Drupal to check the key pair upon init, hence the `poll-rollout` task.

anthony-malkoun: Do you see any issue if we use a HMAC key rather than an RSA key @sonnykt ? It should just be the instructions that change. HMAC will be a single line token that can then be defined at the project level in Lagoon (and overridden for production) and will be able to be managed by our configuration management.

anthony-malkoun:
> > @sonnykt for my understanding why would we ever want to provide env variables for generating keys? Is it good enough to let them be generated every time? The main motivation here is that the `AUTHENTICATED_CONTENT` key has been painful to maintain (Lagoon does not handle multiline config variables correctly) and meant that it had to be added manually for PR environments (and remember to add it for new environments). If there is value to having a known value for the keys, can it be something else like the token that TFA uses or does it have to be a full RSA key?
> 
> If there is no env variable, the module will generate a random key pair upon its installation. The env var is to ensure that all envs have the same key pair.

And more importantly, if I understand correctly @sonnykt that they remain the same after deploy? See my previous comment re: using a HMAC rather than RSA. If this is ok then I'm ok with this. Maintaining the `AUTHENTICATED_CONTENT` key has been a complete pain which I want to avoid.

sonnykt:
> > > @sonnykt for my understanding why would we ever want to provide env variables for generating keys? Is it good enough to let them be generated every time? The main motivation here is that the `AUTHENTICATED_CONTENT` key has been painful to maintain (Lagoon does not handle multiline config variables correctly) and meant that it had to be added manually for PR environments (and remember to add it for new environments). If there is value to having a known value for the keys, can it be something else like the token that TFA uses or does it have to be a full RSA key?
> > 
> > 
> > If there is no env variable, the module will generate a random key pair upon its installation. The env var is to ensure that all envs have the same key pair.
> 
> And more importantly, if I understand correctly @sonnykt that they remain the same after deploy? See my previous comment re: using a HMAC rather than RSA. If this is ok then I'm ok with this. Maintaining the `AUTHENTICATED_CONTENT` key has been a complete pain which I want to avoid.

I don't think OAuth2 server accepts HMAC key. It requires a key pair.

anthony-malkoun:
Thinking some more about this @sonnykt and the install creates a valid key in `public://` which will be persisted across builds. Do you see any issue with just using this initial key ongoing? Because of short comings of environment variables in Lagoon I don't think we will ever need to generate them from environment variables. Thoughts?

sonnykt:
> Thinking some more about this @sonnykt and the install creates a valid key in `public://` which will be persisted across builds. Do you see any issue with just using this initial key ongoing? Because of short comings of environment variables in Lagoon I don't think we will ever need to generate them from environment variables. Thoughts?

The module installation creates a key pair in `private://` (not public) and the key pair remains in the persistent storage. However, once the module is enabled on production and you later launch a new non-prod env, the module does not create a new key pair on that env because it is already installed in the database. The new env doesn't have a valid key pair. The key pair from prod is not copied to the new env, hence the module fails to work.

The `post-rollout` task runs the Drush command to generate the key pair from the env variables, but it doesn't generate a new key pair if the env variables are not set. We could modify that command to do:
* check if there is a valid key pair in `private://`
* if not, try to create a key pair from env variables
* if the env variables are not set, generate a new key pair.

anthony-malkoun:
> > Thinking some more about this @sonnykt and the install creates a valid key in `public://` which will be persisted across builds. Do you see any issue with just using this initial key ongoing? Because of short comings of environment variables in Lagoon I don't think we will ever need to generate them from environment variables. Thoughts?
> 
> The module installation creates a key pair in `private://` (not public) and the key pair remains in the persistent storage. However, once the module is enabled on production and you later launch a new non-prod env, the module does not create a new key pair on that env because it is already installed in the database. The new env doesn't have a valid key pair. The key pair from prod is not copied to the new env, hence the module fails to work.
> 
> The `post-rollout` task runs the Drush command to generate the key pair from the env variables, but it doesn't generate a new key pair if the env variables are not set. We could modify that command to do:
> 
> * check if there is a valid key pair in `private://`
> * if not, try to create a key pair from env variables
> * if the env variables are not set, generate a new key pair.

I think this is the best solution. That post deploy step can stay in place as it won't be destructive. Do you think it needs some sort of `--refresh` or `--force` flag that can get it to generate new key pair, ignoring what's in `private://`. 
It would be nice if Simple OAuth supported the key module (unfortunately not https://www.drupal.org/project/simple_oauth/issues/3133698 - the issue being a short coming in the implementation of the library beneath the module https://github.com/thephpleague/oauth2-server/issues/1007) but that's not an option.

anthony-malkoun:
> The `post-rollout` task runs the Drush command to generate the key pair from the env variables, but it doesn't generate a new key pair if the env variables are not set. We could modify that command to do:
> 
> * check if there is a valid key pair in `private://`
> * if not, try to create a key pair from env variables
> * if the env variables are not set, generate a new key pair.

Hey @sonnykt did this get done?

sonnykt:
> > The `post-rollout` task runs the Drush command to generate the key pair from the env variables, but it doesn't generate a new key pair if the env variables are not set. We could modify that command to do:
> > 
> > * check if there is a valid key pair in `private://`
> > * if not, try to create a key pair from env variables
> > * if the env variables are not set, generate a new key pair.
> 
> Hey @sonnykt did this get done?

I doubt that. IIRC we only discussed but haven't got a final decision yet.

anthony-malkoun:
> > > The `post-rollout` task runs the Drush command to generate the key pair from the env variables, but it doesn't generate a new key pair if the env variables are not set. We could modify that command to do:
> > > 
> > > * check if there is a valid key pair in `private://`
> > > * if not, try to create a key pair from env variables
> > > * if the env variables are not set, generate a new key pair.
> > 
> > 
> > Hey @sonnykt did this get done?
> 
> @anthony-malkoun I doubt that. IIRC we only discussed but haven't got a final decision yet.

Who are you waiting for a final decision from? I'm happy for this to happen.

sonnykt:
@anthony-malkoun you'd probably get @vincent-gao to make the change as my allocation this week is for VT.

anthony-malkoun (approved):

PR #179, opened 2020-09-17
[SDPA-4569] add custom callout template as plugin

rubyllamadora:
#### Changed #### 
- Added custom plugin for callout template added in Media group wysiwyg toolbar

CMS to test: https://nginx-php-content-vic-pr-967.lagoon.vicsdp.amazee.io/
#### Jira #### 
https://digital-engagement.atlassian.net/browse/SDPA-4569
https://digital-engagement.atlassian.net/browse/SDPA-4568
https://digital-engagement.atlassian.net/browse/SDPA-4567


.gitignore
```diff
@@ -10,3 +10,4 @@
 docker-compose.override.yml
 screenshots
 vendor
+.idea
```
anthony-malkoun, line 13: Put this in your global git ignore and revert this line.

config/optional/editor.editor.rich_text.yml
```diff
@@ -42,6 +42,7 @@ settings:
           name: Media
           items:
             - Blockquote
+            - tide_callout
```
anthony-malkoun, line 45: Just a style thing but Camel Case seems to be the norm here?

vincent-gao (approved):

anthony-malkoun (changes requested):
A couple of things that would be nice to have if possible...
1. Where did the image for the new button come from? Something a bit more representative of the function might be good. Could ask UI/UX team what they think should be the button.
2. It populates the with lorum ipsum. Is this as required? I'm thinking some more useful help text perhaps.
3. It doesn't visually style in the CMS at all. At least an indent might help, or the vertical bar preferably
4. Applying this to existing text deletes the text?

config/optional/editor.editor.rich_text.yml
rubyllamadora: I followed `tide_media` presumed how we named custom plugin https://github.com/dpc-sdp/tide_core/pull/179/files#diff-d5400ced7d74da36c387e69173307efbL67

rubyllamadora:
> A couple of things that would be nice to have if possible...
> 
> 1. Where did the image for the new button come from? Something a bit more representative of the function might be good. Could ask UI/UX team what they think should be the button.
> 2. It populates the with lorum ipsum. Is this as required? I'm thinking some more useful help text perhaps.
> 3. It doesn't visually style in the CMS at all. At least an indent might help, or the vertical bar preferably
> 4. Applying this to existing text deletes the text?

Hey @anthony-malkoun, thanks for the review.
1. I created it and didn't spend much time with it as I was hoping the UI/UX will give me a better image when after they review. 
2. Not required, I can update it with useful details. It's just because it needs a placeholder for the custom classes (different styling for title and body text) we added in the template. 
3. Will look into how to improve it visually in CMS.
4. Yes, this actually works as a template but not inside the template dropdown toolbar. Highlighting option will be more complex as they have specific styling for the heading :(  See `__title` and `__content` below.
```
<div class="callout-wrapper">
  <div class="callout-wrapper__title">Lorem ipsum dolor</div>
  <div class="callout-wrapper__content">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</div>
</div>
```

anthony-malkoun:
>     4\. Yes, this actually works as a template but not inside the template dropdown toolbar. Highlighting option will be more complex as they have specific styling for the heading :(  See `__title` and `__content` below.

I don't think this is the case actually. I think the way it will work is they'll style a block of text and just apply the style to a selection. If there are actual styles that are needed for the heading then it probably should be just applied to any, say, `<h2>` elements in the selection. Maybe check with Vanessa or Jo about how this should work.

rubyllamadora:
> > 1. Yes, this actually works as a template but not inside the template dropdown toolbar. Highlighting option will be more complex as they have specific styling for the heading :(  See `__title` and `__content` below.
> 
> I don't think this is the case actually. I think the way it will work is they'll style a block of text and just apply the style to a selection. If there are actual styles that are needed for the heading then it probably should be just applied to any, say, `<h2>` elements in the selection. Maybe check with Vanessa or Jo about how this should work.

Will ask Jo to check the UI. Vanessa has already tested this and find it ok.

anthony-malkoun (approved):
This is great! Thanks for taking the changes on board @rubyllamadora. Approved.

PR #187, opened 2020-10-06
[SW-402] : Update the simple sitemap module

FleetAdmiralButter:


ivangrynenko:
@FleetAdmiralButter 
Any chance this patch can be moved ahead? The module is causing major headaches to sites with a lot of content and the sitemap links count drops to 2000 upon each cron run.

When generating sitemap manually, `drush ssg` gets all content in but then, after a cron run, only the first 2,000 items show up.

I believe version 3 of the sitemap resolved this issue.

PR #197, opened 2020-12-07
[tide_oauth_changes] Updated drush command to generate keys if no envkeys.

MdNadimHossain:
### Changes based on the following command
```
The post-rollout task runs the Drush command to generate the key pair from the env variables, but it doesn't generate a new key pair if the env variables are not set. We could modify that command to do:

check if there is a valid key pair in private://
if not, try to create a key pair from env variables
if the env variables are not set, generate a new key pair.
```

### DIFF
--- a/modules/tide_oauth/drush/tide_oauth.drush.inc
+++ b/modules/tide_oauth/drush/tide_oauth.drush.inc
@@ -24,23 +24,11 @@ function drush_tide_oauth_keygen() {
   /** @var \Drupal\tide_oauth\EnvKeyGenerator $env_key_generator */
   $env_key_generator = \Drupal::service('tide_oauth.env_key_generator');
   // Generate the OAuth encryption keys from Environment variables.
-  if ($env_key_generator->hasEnvKeys()) {
-    if ($env_key_generator->generateEnvKeys()) {
-      // Update Simple OAuth settings.
-      $env_key_generator->setSimpleOauthKeySettings();
-    }
-    else {
-      drush_set_error('Could not generate OAuth keys from TIDE_OAUTH_*_KEY environment variables.');
-    }
+  if ($env_key_generator->generateEnvKeys()) {
+    // Update Simple OAuth settings.
+    $env_key_generator->setSimpleOauthKeySettings();
   }
   else {
-    // Generate keys if Environment variables are not set.
-    if ($env_key_generator->generateOauthKeys()) {
-      // Update Simple OAuth settings.
-      $env_key_generator->setSimpleOauthKeySettings();
-    }
-    else {
-      drush_set_error('Could not generate keys.');
-    }
+    drush_set_error('Could not generate OAuth keys from TIDE_OAUTH_*_KEY environment variables.');
   }
 }
--- a/modules/tide_oauth/drush/tide_oauth.drush.inc
+++ b/modules/tide_oauth/drush/tide_oauth.drush.inc
@@ -11,16 +11,16 @@ use Drupal\tide_oauth\EnvKeyGenerator;
  * Implements hook_drush_command().
  */
 function tide_oauth_drush_command() {
-  $commands['tide-oauth-env-keygen'] = [
+  $commands['tide-oauth-keygen'] = [
     'description' => 'Generate OAuth keys from environment variables',
     'drupal dependencies' => ['tide_oauth'],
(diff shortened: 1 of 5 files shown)


anthony-malkoun (changes requested):
Hey @MdNadimHossain I think you need to do a similar change to https://github.com/dpc-sdp/tide_core/blob/feature/tide_oauth/modules/tide_oauth/src/Commands/TideOauthCommands.php so that it supports later versions of drush as well.

Also, let's just rename the function `drush_tide_oauth_keygen` maybe as it is now more general purpose.

anthony-malkoun (approved):
👍

modules/tide_oauth/drush/tide_oauth.drush.inc
```diff
@@ -33,4 +37,15 @@ function drush_tide_oauth_env_keygen() {
       drush_set_error('Could not generate OAuth keys from TIDE_OAUTH_*_KEY environment variables.');
     }
   }
+  else {
+    /** @var \Drupal\simple_oauth\Service\KeyGeneratorService $key_generator */
+    $key_generator = \Drupal::service('simple_oauth.key.generator');
```
sonnykt, line 42: @MdNadimHossain @anthony-malkoun I think those lines should come as a public method of `tide_oauth.env_key_generator` service to reduce the duplication in both Drush 8 and Drush 9 commands and also to keep the Drush command minimal.

Probably just `$env_key_generator->generateKeys()` for the logic of the whole command:
* check for `$this->hasEnvKeys()` and call `$this->generateEnvKeys()`
* otherwise call another method to generate keys via `$key_generator->generateKeys('private://');`.

sonnykt (changes requested):

modules/tide_oauth/drush/tide_oauth.drush.inc
MdNadimHossain: Thanks, @sonnykt I have cleaned up the code 👍

modules/tide_oauth/src/Commands/TideOauthCommands.php
```diff
@@ -33,23 +33,21 @@ public function __construct(EnvKeyGenerator $env_key_generator) {
...
-        $this->envKeyGenerator->setSimpleOauthKeySettings();
-        $this->io()->success('OAuth keys have been created from TIDE_OAUTH_*_KEY environment variables.');
-      }
-      else {
-        $this->io()->error('Could not generate OAuth keys from TIDE_OAUTH_*_KEY environment variables.');
-      }
+    if ($this->envKeyGenerator->generateEnvKeys()) {
+      // Update Simple OAuth settings.
+      $this->envKeyGenerator->setSimpleOauthKeySettings();
+      $this->io()->success('OAuth keys have been created from TIDE_OAUTH_*_KEY environment variables.');
```
sonnykt, line 47: @MdNadimHossain @anthony-malkoun this info message may be misleading now, given that the keys may be generated not from the env.

MdNadimHossain: @sonnykt should I give a more general one like this - "OAuth keys have been created" ?

modules/tide_oauth/src/EnvKeyGenerator.php
```diff
@@ -89,7 +100,7 @@ public function hasEnvKeys() : bool {
    */
   public function generateEnvKeys() : bool {
     if (!$this->hasEnvKeys()) {
-      return FALSE;
+      $this->generateOauthKeys();
```
sonnykt, line 103: @MdNadimHossain should it return right here? `$this->generateOauthKeys()` generates the keypair, and the subsequent lines overwrite the keypair with values of the env variables -  which do not exist.

sonnykt (changes requested):

modules/tide_oauth/src/EnvKeyGenerator.php
MdNadimHossain: @sonnykt oh yes it shouldn't return. It should just create the oauth.key and oauth.pub if  EnvKeys doesn't exist. Then it will continue to create the private and public key right!

MdNadimHossain: @sonnykt Updated thanks

sonnykt (approved):

PR #208, opened 2021-01-07
Adds enhanced_mime_type_filter filter

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-3398

### Changes
1.  add a custom view filter.

image


MdNadimHossain (commented):
Hey, @vincent-gao is it possible to create a custom view like tide_files and disable the current one which is files. Then make the changes on the custom one, rather than changing the core one. The reason that I say this is because we have done the same thing for other views. For example tide_media_browser or tide_site_moderated_content_filters.

vincent-gao:
> Hey, @vincent-gao is it possible to create a custom view like tide_files and disable the current one which is files. Then make the changes on the custom one, rather than changing the core one. The reason that I say this is because we have done the same thing for other views. For example tide_media_browser or tide_site_moderated_content_filters.

em , thanks @MdNadimHossain , yeah , replace the older view with a custom view sounds good.

MdNadimHossain (commented):
@vincent-gao Sorry I should have mentioned it in earlier CR, but I think it will be better if we follow the naming like `tide_enhanced_files`, so that it matches with our other custom ones and easily identifiable

vincent-gao:
> tide_enhanced_files

no worries @MdNadimHossain , how about `tide_enhanced_mime_type_filter` you happy with this name?

MdNadimHossain:
> > tide_enhanced_files
> 
> no worries @MdNadimHossain , how about `tide_enhanced_mime_type_filter` you happy with this name?

:D anything with tide sounds fine. Sorry for causing the pain 😞

MdNadimHossain (approved):

PR #210, opened 2021-01-11
[SDPSUP-2516] Added code to prepare the private directory before generating oauth keys.

MdNadimHossain:
### Jira
https://digital-engagement.atlassian.net/browse/SDPSUP-2516

### Issue
During openshift build, the post-rollout tasks run the drush command to generate the oauth keys [here](https://github.com/dpc-sdp/content-vic-gov-au/blob/develop/.lagoon.yml#L41)
it throws the error saying - 
```
Drupal\simple_oauth\Service\Exception\FilesystemValidationException:     [error]
--
  | Directory "private://" is not a valid directory. in /app/docroot/modules/contrib/simple_oauth/src/Service/Filesystem/FilesystemValidator.php:54
```
The function `generateOauthKeys()` needs to prepare the private directory as well just like how it is done in `generateEnvKeys()`

### DIFF
--- a/modules/tide_oauth/src/EnvKeyGenerator.php
+++ b/modules/tide_oauth/src/EnvKeyGenerator.php
@@ -165,6 +165,9 @@ class EnvKeyGenerator {
       return FALSE;
     }
 
+    $private = 'private://';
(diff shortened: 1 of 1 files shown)


anthony-malkoun (approved):
Looks good to me.

PR #212, opened 2021-01-22
[SDPA-4946] Fixed indentation errors.

GROwen:
```
In YamlPecl.php line 74:

  yaml_parse(): parsing error encountered during parsing: did not find expect
  ed key (line 5, column 7), context while parsing a block mapping (line 3, c
  olumn 5)
```
On upgrade of Drush the above error was thrown for any Drush commands.

Cause was extra indents in the updated file


anthony-malkoun (approved):
Ah, good pickup!

MdNadimHossain (approved):

PR #213, opened 2021-01-22
Moved Topic and Tags field to node form sidebar and Added Show Intro text field.

sonnykt:


tide_core.module
```diff
@@ -263,6 +264,61 @@ function tide_core_editor_js_settings_alter(array &$settings) {
...
+ * @param array $form
+ *   The complete form structure.
+ *
+ * @return array
+ *   The processed form.
+ */
+function _tide_core_form_node_form_process(array $element, FormStateInterface $form_state, array $form = []) {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $form_state->getFormObject()->getEntity();
+  $element['tide_core_metadata'] = [
```
MdNadimHossain, line 293: That is a nice way to do it. I was thinking of doing it through the field_group option, but this way is much easier and cleaner. :+1:

MdNadimHossain (approved):

PR #216, opened 2021-04-14
Added package/ckeditor-fakeobjects in repositories

gargsuchi:
When trying to install tide locally, composer gives this error:
- dpc-sdp/tide_core 2.0.4 requires ckeditor/fakeobjects ^4.5.11 -> no matching package found

This PR fixes the error.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,6 +70,22 @@
         "drupal": {
             "type": "composer",
             "url": "https://packages.drupal.org/8"
+        },
+        "package/ckeditor-fakeobjects": {
+            "type": "package",
+            "package": {
+                "name": "ckeditor/fakeobjects",
+                "type": "drupal-library",
+                "version": "4.5.11",
+                "dist": {
+                    "type": "zip",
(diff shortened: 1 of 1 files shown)


barbun:
@gargsuchi good stuff :) 

It seems that the PR pins the package version to `4.5.11`, all while `composer.json` constrains it only to non-breaking updates `^4.5.11`. I am not sure how critical this is but it is probably a good practice to stick to non-breaking updates, especially when other packages are also set to use version range.

From here, I guess there's at least 2 options:
- Lock the version in `composer.json` to `4.5.11`
- Switch to `"drupal-ckeditor-libraries-group/fakeobjects": "^4.5.11"` which seems to be mirroring CKEditor updates.

barbun:
Enhanced on this in #219. I reckon this PR can be closed then :)

gargsuchi:
Closing in favour of https://github.com/dpc-sdp/tide_core/pull/229 @barbun

PR #221, opened 2021-04-22
remove xss patch

vincent-gao:
###  Motivation
1. Drupalcore releases the latest version that has the patch applied , so no need `Refactor Xss::attributes() to allow filtering of style attribute values - https://www.drupal.org/project/drupal/issues/3109650`.
2. without removing , a fatal error occurs during `composer update`.


anthony-malkoun (approved):

vincent-gao:
sorry, my mistake, the patch should stay.

PR #226, opened 2021-04-23
Only enabled tide_edit_protection if its role dependency is met.

sonnykt:


vincent-gao (approved):
Thanks @sonnykt !

PR #231, opened 2021-05-04
Remove the install hook.

vincent-gao:
### Changes
1. the module should get installed from baywatch module

https://github.com/dpc-sdp/baywatch/pull/24

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -1082,3 +1082,10 @@ function tide_core_update_8045() {
     user_role_grant_permissions(Role::load($role)->id(), $permissions);
   }
 }
+
+/**
+ * Enable tide_edit_protection module.
+ */
+function tide_core_update_8046() {
+  // @todo This update will be removed before Drupal 9.0.0
+}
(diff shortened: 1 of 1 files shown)


tide_core.install
```diff
@@ -1082,15 +1082,3 @@ function tide_core_update_8045() {
     user_role_grant_permissions(Role::load($role)->id(), $permissions);
   }
 }
-
-/**
- * Enable tide_edit_protection module.
- */
-function tide_core_update_8046() {
```
sonnykt, line 1089: @vincent-gao this update hook was released in 2.0.8 so you can't remove it. The body of the hook can be removed. but the hook itself must stay.

sonnykt (changes requested):

tide_core.install
vincent-gao: @sonnykt  sure. indeed it should keep same number with database.

sonnykt (approved):

PR #232, opened 2021-05-04
Add tide_ui_restriction module

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-5122

### Motivation
1. as an open source project, we don't want the changes that will affect non-DPC users.
2. we also can provide further logic to restrict more routes, which will not affect non-DPC users.

### Changes
1. adds tide_ui_restriction custom module that will responsible for blocking all users accessing module_list view.

*Please leave some thoughts and comments, thanks*


sonnykt (commented):
@vincent-gao the test `tests/behat/features/modules_management.feature` could stay, you only need to tag it with `@install:tide_ui_restriction`. See https://github.com/dpc-sdp/dev-tools/blob/develop/tests/behat/bootstrap/TideExtensionsTrait.php

sonnykt (approved):

PR #237, opened 2021-05-25
Updates EnvKeyGenerator method

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-5184

### Changes
1. passes 3rd parameter `FileSystemInterface::EXISTS_REPLACE` to `move()` method.


anthony-malkoun (changes requested):
If this fixes the problem then that's a good find! Please check though that permissions are right on a clean build in Lagoon before merging.

vincent-gao:
> If this fixes the problem then that's a good find! Please check though that permissions are right on a clean build in Lagoon before merging.

I think this only fixes `_0`,`_1` `_02` ... suffix . but, yes, I will try it on lagoon env.

vincent-gao:
Hey @anthony-malkoun 

in lagoon env, it is still in 777 permissions. 
after regenerated key pair
image
still in 777.
So we need help from devops? or Steve?

PR #240, opened 2021-06-02
[SDPA-5210] Refactors tide_lock_content view

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-5210

### Changes
1. adds new permission
2. refactor `tide_lock_content`

image

### DIFF
--- a/modules/tide_edit_protection/tide_edit_protection.install
+++ b/modules/tide_edit_protection/tide_edit_protection.install
@@ -52,7 +52,7 @@ function tide_edit_protection_update_8001() {
  * Adds node site filter to lock_content view.
  */
 function tide_edit_protection_update_8002() {
-  if (!\Drupal::moduleHandler()->moduleExists('tide_site_restriction')){
+  if (!\Drupal::moduleHandler()->moduleExists('tide_site_restriction')) {
     return;
   }
   $site_filter =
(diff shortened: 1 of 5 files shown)


modules/tide_edit_protection/tide_edit_protection.install
```diff
@@ -46,3 +47,95 @@ function tide_edit_protection_update_8001() {
...
+      'vid' => 'sites',
+      'hierarchy' => TRUE,
+      'error_message' => TRUE,
+      'plugin_id' => 'taxonomy_index_tid',
+    ];
+
+  $config = \Drupal::configFactory()
+    ->getEditable('views.view.tide_locked_content')
+    ->get('display');
+  if (!isset($config['default']['display_options']['filters']['field_node_site_target_id'])) {
```
yeniatencio, line 123: If you are verifying 'field_node_site_target_id' exists why don't you verify default, display_options and filters as well? Just I am wondering that.

yeniatencio (approved):
Looks good to me.

modules/tide_edit_protection/tide_edit_protection.install
vincent-gao: yeah, in this case, the logic behind,  the only thing that I should care about is just whether  `field_node_site_target_id` exist or not, as I assume that the clients who enabled the module should have the view.  if it doesn't, it is not our responsibility.

PR #244, opened 2021-06-08
Implement scaffold and fix reference build

anthony-malkoun:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/DDS-808

### Changed

1.  Add missing dependencies for install
2. Add checking on CKEditor config update so that config doesn't get corrupted.

### DIFF
--- a/tide_core.info.yml
+++ b/tide_core.info.yml
@@ -48,6 +48,7 @@ dependencies:
   - events_log_track:event_log_track_workflows
   - fakeobjects:fakeobjects
   - fast_404:fast404
+  - link_field_autocomplete_filter:link_field_autocomplete_filter
   - linkit:linkit
   - login_redirect_per_role:login_redirect_per_role
   - m4032404:m4032404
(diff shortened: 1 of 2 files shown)


tide_core.info.yml
```diff
@@ -62,9 +63,9 @@ dependencies:
   - pathauto:pathauto
   - redirect:redirect
   - role_delegation:role_delegation
-  - simple_sitemap:simple_sitemap
```
anthony-malkoun, line 65: This was duplicated two lines below.

tide_core.install
```diff
@@ -945,6 +945,9 @@ function tide_core_update_8038() {
   foreach ($configs as $config) {
     $editable_config = $config_factory->getEditable($config);
     $rows = $editable_config->get('settings.toolbar.rows.0');
+    if (!isset($rows)) {
```
anthony-malkoun, line 948: On a fresh install the config factory could not access the active configuration. Escaping early seems to be fine as the correct config is being installed for a fresh install.

anthony-malkoun:
@FleetAdmiralButter I left the check in there as it's not a terrible check to have but I've excluded the problematic hooks as well.

FleetAdmiralButter (approved):

PR #251, opened 2021-06-28
Fixes redirect issues when deletion happens on CLI

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-5277

### Motivation
content_lock prevents content from being deleted by migrate-rollback, and `drupal ed` or a custom drush command.

### Changes
~1. implements an event subscriber which will release locked contents for migrate-rollback.~
~2. apply https://www.drupal.org/project/drupal/issues/3065207#comment-13214627 patch to fix CLI mode issue.
    - CLI commands will bypass content_lock logic.~

---

1. Adds predelete hook to tide_edit_protection module.
2. Moves the predelete hook  that content_lock defined to the end of the running order.


anthony-malkoun:
@sonnykt can you take a look at this for me please. My concern is that it fixes the migration rollback issue but doesn't fix say, tide_demo_content being uninstalled.

vincent-gao:
> @sonnykt can you take a look at this for me please. My concern is that it fixes the migration rollback issue but doesn't fix say, tide_demo_content being uninstalled.

@anthony-malkoun sorry, I didn't mention it on description. it applies https://www.drupal.org/project/drupal/issues/3065207#comment-13214627 patch which fixes uninstalling tide_demo_content issue.

sonnykt:
@vincent-gao the event subscriber is only applicable if the site has Migrate module enabled. Not all sites have it. The patch from Drupal core for `forceDelete()` is for Drupal 9 and we are not there yet, I'm not sure what is the impact on other contrib modules as I see it as a core API change.

imho the `predelete()` hook in content_lock module is critically wrong. Applying its patch doesn't really fix. The module is doing the wrong action to stop the entity deletion. I think we should implement a predelete hook that runs  to release the lock before content_lock kicks in.

vincent-gao:
> @vincent-gao the event subscriber is only applicable if the site has Migrate module enabled. Not all sites have it. The patch from Drupal core for `forceDelete()` is for Drupal 9 and we are not there yet, I'm not sure what is the impact on other contrib modules as I see it as a core API change.
> 
> imho the `predelete()` hook in content_lock module is critically wrong. Applying its patch doesn't really fix. The module is doing the wrong action to stop the entity deletion. I think we should implement a predelete hook that releases the lock before content_lock kicks in.

Oh sorry , I just posted wrong link 
it should be https://www.drupal.org/project/content_lock/issues/2951652#comment-12751913
and ok will do predeletehook way.

sonnykt:
@vincent-gao see the comment #9 on the issue of content_lock https://www.drupal.org/project/content_lock/issues/2951652#comment-12829088

modules/tide_edit_protection/tide_edit_protection.module
```diff
@@ -142,3 +143,36 @@ function tide_edit_protection_preprocess_status_messages(&$variables) {
     }
   }
 }
+
+/**
+ * Implements hook_entity_predelete().
+ */
+function tide_edit_protection_entity_predelete(Drupal\Core\Entity\EntityInterface $entity) {
```
sonnykt, line 150: @vincent-gao please don't use full qualified namespace in parameters.

modules/tide_edit_protection/tide_edit_protection.module
```diff
@@ -142,3 +143,38 @@ function tide_edit_protection_preprocess_status_messages(&$variables) {
...
+  if (PHP_SAPI == 'cli') {
```
sonnykt, line 151: the predelete hook of content_lock is totally wrong and it causes more issue upon entity deletion not just via CLI. Hence we should release the lock regardless of the situation.
```php
$lock_service->release($entity_id, LanguageInterface::LANGCODE_NOT_SPECIFIED);
```
Just this line is enough. No need to check for CLI, no need to fetch the lock.

vincent-gao:
> @vincent-gao see the comment #9 on the issue of content_lock https://www.drupal.org/project/content_lock/issues/2951652#comment-12829088

Just made a change, please take a look @sonnykt

modules/tide_edit_protection/tide_edit_protection.module
```diff
@@ -142,3 +143,38 @@ function tide_edit_protection_preprocess_status_messages(&$variables) {
...
+      $lock_service->release($entity_id, LanguageInterface::LANGCODE_NOT_SPECIFIED);
+    }
+  }
+}
+
+/**
+ * Implements hook_module_implements_alter().
+ */
+function tide_edit_protection_module_implements_alter(&$implementations, $hook) {
+  if (PHP_SAPI == 'cli') {
```
sonnykt, line 170: @vincent-gao just increase module_weight of `content_lock` to a big number (eg. 1000) using an update hook in tide_edit_protection. No need to implement this hook.

sonnykt (changes requested):

sonnykt (approved):

modules/tide_edit_protection/tide_edit_protection.module
```diff
@@ -142,3 +144,11 @@ function tide_edit_protection_preprocess_status_messages(&$variables) {
     }
   }
 }
+
+/**
+ * Implements hook_entity_predelete().
+ */
+function tide_edit_protection_entity_predelete(EntityInterface $entity) {
+  \Drupal::service('content_lock')
```
sonnykt, line 152: @vincent-gao imo we should have some comments indicating why we need this hook, probably referencing the Drupal issues.

anthony-malkoun (approved):
Thank you for feedback @sonnykt. This seems like a much cleaner solution to me now. Approved.

anthony-malkoun:
The only thought I had is does this make it now possible to delete locked content by an editor? Is this even something to be worried about as I don't think many users have the permission to delete and could probably break the lock anyway.

sonnykt:
@anthony-malkoun if the purpose of `tide_edit_protection` is to prevent content from being deleted by editors, `content_lock` is not really the right choice. I would expect the logic to be more complex with permissions and site restrictions. If the sole purpose is to prevent concurrent editing, releasing the lock upon deletion should not have any issue.

anthony-malkoun:
> @anthony-malkoun if the purpose of `tide_edit_protection` is to prevent content from being deleted by editors, `content_lock` is not really the right choice. I would expect the logic to be more complex with permissions and site restrictions. If the sole purpose is to prevent concurrent editing, releasing the lock upon deletion should not have any issue.

Yep, agree. I think it's really only there as a guardrail rather than a hard limit. If you have permissions to delete you have permissions to break the lock anyway.

sonnykt:
hi @anthony-malkoun @vincent-gao I think we should release this change asap. On env with PERSIST_DB on, during deployment the `drush cim` will uninstall tide_demo_content and it is guaranteed to break.

vincent-gao:
> hi @anthony-malkoun @vincent-gao I think we should release this change asap. On env with PERSIST_DB on, during deployment the `drush cim` will uninstall tide_demo_content and it is guaranteed to break.

sure @sonnykt , I am about to release it right away.

PR #254, opened 2021-07-13
[DDS-245] Fix wrong timezone is configured after a site install

mayurngondhkar:
Jira: https://digital-engagement.atlassian.net/browse/DDS-245

1. Add system.date.yml (copied from content VT)
image

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -53,6 +53,12 @@ function tide_core_install() {
     ->set('register', USER_REGISTER_ADMINISTRATORS_ONLY)
     ->save(TRUE);
 
+  // Change timezone to Australia/Melbourne.
+  \Drupal::configFactory()
+    ->getEditable('system.date')
+    ->set('timezone.default', 'Australia/Melbourne')
+    ->save(TRUE);
+
   // Creates terms for Topic vocabulary.
   _tide_core_create_topic_terms();
 


MdNadimHossain (changes requested):
@mayurngondhkar may be adding a behat test will be good as well

anthony-malkoun:
> @mayurngondhkar may be adding a behat test will be good as well

Yes please. I would like all of these changes to have behat tests with them. To test we can run my draft branch of the reference site which will install a site from scratch.

anthony-malkoun:
@mayurngondhkar / @MdNadimHossain I've been thinking about this. Should it actually live in `tide_core`? This is probably something that should be in `baywatch`, that is something that is for SDP only rather than all users of `tide`.

CC: @vincent-gao

MdNadimHossain:
> @mayurngondhkar / @MdNadimHossain I've been thinking about this. Should it actually live in `tide_core`? This is probably something that should be in `baywatch`, that is something that is for SDP only rather than all users of `tide`.
> 
> CC: @vincent-gao

@anthony-malkoun hmm is that because the timezone defaulted to Melbourne only applies to the SDP clients and tide_core is open source and if other organization uses it, they might not want it to set to Melbourne time zone?

vincent-gao:
> @mayurngondhkar / @MdNadimHossain I've been thinking about this. Should it actually live in `tide_core`? This is probably something that should be in `baywatch`, that is something that is for SDP only rather than all users of `tide`.
> 
> CC: @vincent-gao

Yes, agreed. non-SDP user  (NON Victoria user ⏲️ ) should able to change timezone settings.

mayurngondhkar:
@anthony-malkoun True. As @MdNadimHossain mentioned being open source it should not be defaulted to a given timezone.

anthony-malkoun:
I'm going to close this in favour of https://github.com/dpc-sdp/sdp-ansible-upgrade/pull/83 even though it's a WIP as it makes more sense for it to go there.

PR #256, opened 2021-07-20
Fixed update hooks with incorrect use of $config_factory->getEditable().

sonnykt:
Fresh install is broken at may places due to incorrect use of `$config_factory->getEditable()` in update hooks.

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -416,6 +416,9 @@ function tide_core_update_8015() {
   if (\Drupal::moduleHandler()->moduleExists('clamav')) {
     $config_factory = \Drupal::configFactory();
     $config = $config_factory->getEditable('clamav.settings');
+    if ($config->isNew()) {
+      return;
+    }
     $config->set('scan_mode', 0);
     $config->set('mode_daemon_tcpip.hostname', 'clamav.sdp-central-clamav-master.svc.cluster.local');
     $config->save();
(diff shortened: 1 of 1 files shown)


MdNadimHossain:
@sonnykt does the update_hooks run during a fresh install? My understanding was, only the install hook run during the fresh install and update hooks only runs for the existing project where the module is enabled :/

sonnykt:
@MdNadimHossain we have this in `tide_core_install()` and it becomes troublesome:
```
$excluded = [
    'tide_core_update_8015',
    'tide_core_update_8038',
    'tide_core_update_8043',
  ];
  $functions = get_defined_functions();
  foreach ($functions['user'] as $function) {
    if (strpos($function, 'tide_core_update_') === 0) {
      // We don't want the tide_core_update_8015 to run during the CI.
      if (in_array($function, $excluded)) {
        continue;
      }
      call_user_func($function);
    }
  }```

anthony-malkoun:
So, to be clear, normally an install _doesn't_ run update hooks but in the case of `tide_core` we're manually forcing a run of (most of) the update hooks. Sound right?

vincent-gao:
> So, to be clear, normally an install _doesn't_ run update hooks but in the case of `tide_core` we're manually forcing a run of (most of) the update hooks. Sound right?

 Yes.

sonnykt:
> So, to be clear, normally an install _doesn't_ run update hooks but in the case of `tide_core` we're manually forcing a run of (most of) the update hooks. Sound right?

yes, it's been there since inception due to the laziness of developers 😂

anthony-malkoun (approved):

PR #264, opened 2021-08-27
Add block_inactive_users module

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-5438

### Changes
1. adds two drush commands
      - `inactive-notify` and `inactive-block`
2. The drush command run every 1 hour.
3.  usage, 2 ways.
      -   updates `.lagoon.yml` to have below
    ```
      - name: inactive users - notify
        schedule: "0 * * * *"
        command: 'drush inactive-notify'
        service: cli
      - name: inactive users - block
        schedule: "0 * * * *"
        command: 'drush inactive-block'
        service: cli
    ```
    -  running by  core `cron`
### Issues with `block_inactive_users` module
1. `block_inactive_users` module send emails and blocks users at same time. (This is not we want.)
2.  `block_inactive_users` bypass the queue_mail by sending email directly in php without using mail hook.
3.  we want to send notifications first, and after 1 month, we block users who still didn't login after got the notification.


MdNadimHossain (commented):
@vincent-gao Looks good to me. But it will create issues for vicpol. Because, vicpol has this media role attached to the users and these users never logs in. The mail service just check the users with this role and if they are active it sends out the email. 
So if they use this module, it will block the users with media role and they won't get the media release email. So, should there be an option to exclude a specific role or should this be communicated to someone. Can you please confirm.

PR #270, opened 2021-09-28
[SDPA-5400] : Don't redirect to Dashboard on OAuth routes

FleetAdmiralButter:


modules/tide_oauth/tide_oauth.module
```diff
@@ -69,3 +69,13 @@ function tide_oauth_form_simple_oauth_authorize_form_alter(&$form, FormStateInte
...
+
+/**
+ * Implements hook_module_implements_alter().
+ */
+function tide_oauth_module_implements_alter(&$implementations, $hook) {
+  $destination = \Drupal::service('redirect.destination')->get();
+  if (strpos($destination, '/oauth/authorize') !== false) {
+    unset($implementations['login_redirect_per_role']);
+  }
+}
```
MdNadimHossain, line 81: Looks good to me. Might need to put one line gap at the end to pass the linting though in ci

MdNadimHossain (commented):
Just one linting issue, everything else looks fine. Good find

modules/tide_dashboard/tide_dashboard.module
```diff
@@ -29,3 +29,14 @@ function tide_dashboard_toolbar_alter(&$items) {
 function _tide_dashboard_workbench_content_title_callback() {
   return t('Dashboard');
 }
+
+/**
+ * Implements hook_user_login().
+ */
+function tide_dashboard_user_login($account) {
+  $destination = \Drupal::service('redirect.destination')->get();
+  if (!(strpos($destination, '/oauth/authorize') !== FALSE)) {
```
FleetAdmiralButter, line 38: Not sure if these routes should be abstracted into config - they're quite unlikely to change.

tide_core.info.yml
```diff
@@ -50,7 +50,6 @@ dependencies:
   - fast_404:fast404
   - link_field_autocomplete_filter:link_field_autocomplete_filter
   - linkit:linkit
-  - login_redirect_per_role:login_redirect_per_role
```
FleetAdmiralButter, line 53: I don't think we need `login_redirect_per_role` - it conflicts with Simple OAuth and breaks preview redirects after user login. All roles also redirect to `/admin/workbench` after login so we're not using the module's main feature anyway.

MdNadimHossain (approved):
Looks good to me now

PR #277, opened 2021-11-08
[SDPA-5674] Tide JIRA module

FleetAdmiralButter:
JIRA ticket: https://digital-engagement.atlassian.net/browse/SDPA-5815

1. This PR adds a new submodule to tide_core called tide_jira.
2. tide_jira detects when content enters the needs_review or archived_pending states, and derives metadata from the node revision, such as author, revision log message, preview links and time.
3. This information is turned into a data model, TideJiraTicketModel, and is added to a queue called "tide_jira".
4. During Drupal cron, this queue is processed by TideJiraProcessorBase, which reads items off the queue and creates a tickets using the JIRA Rest API.

Installation of this module does not require any additional changes, as the JIRA credentials are already injected as environment variables. However, it can only be used on sites with the Department field, such as content.vic.gov.au.


composer.json
```diff
@@ -75,7 +75,9 @@
         "drupal/inline_entity_form": "^1.0",
         "drupal/jquery_ui_touch_punch": "^1.0",
         "drupal/admin_audit_trail": "^1.0",
-        "drupal/block_inactive_users": "^2.0"
+        "drupal/block_inactive_users": "^2.0",
+        "lesstif/php-jira-rest-client": "2.6.0",
```
MdNadimHossain, line 79: just a note, if we update to php 8 then will have to upgrade this to 3.0.0

modules/tide_jira/README.md
```diff
@@ -0,0 +1,46 @@
...
+* A general configuration page is available at /admin/config/development/tide_jira.
+* During installation, a default JIRA endpoint is configured. The username and password are retrieved from the environment variables `JIRA_USERNAME` and `JIRA_PASSWORD`. Please make sure these are configured in the PHP environment.
+
+
+MAINTAINERS
+-----------
+
+Current maintainers:
+* Single Digital Presence -
+  https://github.com/dpc-sdp
```
MdNadimHossain, line 46: Love this documentation :heart_eyes:

modules/tide_jira/src/TideJiraAPI.php
```diff
@@ -0,0 +1,290 @@
...
+   * Returns field_jira_project for a taxonomy.
+   *
+   * @param int $tid
+   *   Taxonomy term ID.
+   *
+   * @return string
+   *   Returns value of field_jira_project.
+   */
+  private function getProjectInfo($tid) {
+    $dept = $this->entityTypeManager->getStorage('taxonomy_term')->load($tid);
```
MdNadimHossain, line 158: I think you need to check the $tid existence here as well. What it passes the NULL value

modules/tide_jira/tide_jira.module
```diff
@@ -0,0 +1,26 @@
...
+ * Contains tide_jira.module.
+ */
+
+use Drupal\node\NodeInterface;
+
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ */
+function tide_jira_node_presave(NodeInterface $node) {
+  // Never let an error working with JIRA stop content from being saved.
```
MdNadimHossain, line 14: nice

modules/tide_jira/tide_jira.info.yml
```diff
@@ -0,0 +1,7 @@
+name: Tide Jira
+type: module
+description: Tide Jira integration
+core_version_requirement: ^8.9 || ^9
+package: tide
+dependencies:
+  - jira_rest:jira_rest
```
MdNadimHossain, line 7: I think this one needs to be added in the dependencies as well - lesstif/php-jira-rest-client. Also the related fields I guess

MdNadimHossain (changes requested):
@FleetAdmiralButter overall looks good except for few small thing. But the main concern is, this on looks like more content-vic specific changes if I am not wrong. Because looking into the department field the field storage belongs to content-vic. Also some other details about Jira settings which looks content-vic specific for me. So need to confirm those things.

modules/tide_jira/config/install/field.field.taxonomy_term.department.field_jira_project.yml
```diff
@@ -0,0 +1,18 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.taxonomy_term.field_jira_project
+    - taxonomy.vocabulary.department
```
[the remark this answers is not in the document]
FleetAdmiralButter: Hey @MdNadimHossain, this module will only ever be used on content-vic since it is the only site that uses the Federated Publishing model.

modules/tide_jira/config/install/tide_jira.settings.yml
```diff
@@ -0,0 +1,4 @@
+customer_request_type_id: 37e7a69b-3d10-411e-877b-43dc66413613
+customer_request_type_field_id: customfield_10025
+no_account_email: nojiraaccount@dpc.vic.gov.au
+issue_type: 'Service Request'
```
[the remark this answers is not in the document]
FleetAdmiralButter: These field IDs are specific to our JIRA instance. They should not change with time but if we ever move to a new instance they will need to be updated.

modules/tide_jira/tide_jira.info.yml
FleetAdmiralButter: lesstif/php-jira-rest-client is a PHP library, not a module, I don't think it can be specified here?

modules/tide_jira/src/TideJiraAPI.php
```diff
@@ -217,7 +222,7 @@ private function getAuthorInfo(NodeInterface $node) {
         'account_id' => '',
         'name' => $node->getRevisionUser()->get('name')->value . ' ' . $node->getRevisionUser()->get('field_last_name')->value,
         'department' => $this->entityTypeManager->getStorage('taxonomy_term')->load($node->getRevisionUser()->get('field_department_agency')->first()->getValue()['target_id'])->getName(),
-        'project' => $this->getProjectInfo($node->getRevisionUser()->get('field_department_agency')->first() ? $node->getRevisionUser()->get('field_department_agency')->first()->getValue()['target_id'] : NULL),
+        'project' => $node->getRevisionUser()->get('field_department_agency')->first() ? $this->getProjectInfo($node->getRevisionUser()->get('field_department_agency')->first()->getValue()['target_id']) : NULL,
```
FleetAdmiralButter, line 225: Fixed this call so that it will never pass `NULL` to getProjectInfo()

MdNadimHossain (approved):
LGTM now
Thanks @FleetAdmiralButter for making all the changes :+1:

PR #285, opened 2021-12-10
[SDPA-5923] Remove permissions_by_term module

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-5923

### Change
remove permissions_by_term out of from tide_core.

### DIFF
--- a/tide_core.module
+++ b/tide_core.module
@@ -368,21 +368,18 @@ function tide_core_menu_link_content_access(EntityInterface $menu_link_item, $op
  * Implements hook_node_access_records().
  */
 function tide_core_node_access_records(NodeInterface $node) {
-  // Only run if the module permission by terms is enabled.
-  if (\Drupal::moduleHandler()->moduleExists('permissions_by_term')) {
-    if (!$node->isPublished()) {
-      $grants = [];
-      $grants[] = [
-        'realm' => 'tide_core',
-        'gid' => 1,
-        'grant_view' => 1,
-        'grant_update' => 0,
-        'grant_delete' => 0,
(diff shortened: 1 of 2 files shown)


tide_core.module
```diff
@@ -412,13 +406,11 @@ function tide_core_node_grants(AccountInterface $account, $op) {
  */
 function tide_core_node_access(NodeInterface $node, $op, AccountInterface $account) {
```
anthony-malkoun, line 407: @vincent-gao if this was here because of PBT, should the whole hook go?

anthony-malkoun (changes requested):
Vincent, should the hooks be removed totally? Do they do anything besides work with permissions by term? What about sites that still have authenticated content?

tide_core.module
vincent-gao: hey @anthony-malkoun . this was for https://digital-engagement.atlassian.net/browse/SRM-24 , but because of PBT.

vincent-gao:
> Vincent, should the hooks be removed totally? Do they do anything besides work with permissions by term? What about sites that still have authenticated content?

hey @anthony-malkoun , the hooks are for the dashboard.  the sites, like SSP and VICPOL, will keep permissions_by_term module.

tide_core.module
anthony-malkoun: So this hook is required for sites that don't have PBT? I'm just not sure what we're trying to achieve here.

anthony-malkoun (approved):
Thanks for explaining Vincent.

PR #288, opened 2021-12-20
[SDPA-5931] Fixed "Log in now" link for CMS redirection

edyuenyw:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-5931

### Changed

1. Checked and set URL redirect destination when available.

### Screenshots
Example url part to use: **/?destination=/admin/config/content/embed**

image

**BEFORE**
SDPA-5931-Not-Redirected

**AFTER**
image

### DIFF
--- a/modules/tide_dashboard/tide_dashboard.module
+++ b/modules/tide_dashboard/tide_dashboard.module
@@ -37,6 +37,8 @@ function tide_dashboard_user_login($account) {
   $destination = \Drupal::service('redirect.destination')->get();
   if (!(strpos($destination, '/oauth/authorize') !== FALSE)) {
     $request = \Drupal::service('request_stack')->getCurrentRequest();
-    $request->query->set('destination', '/admin/workbench');
+    if (($request->query->has('destination')) === FALSE) {
+      $request->query->set('destination', '/admin/workbench');
+    }
   }
 }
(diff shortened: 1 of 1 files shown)


vincent-gao (approved):

anthony-malkoun:
@FleetAdmiralButter do you mind doing a quick review of this? What was the original issue you were trying to fix? This shouldn't affect it?

modules/tide_dashboard/tide_dashboard.module
```diff
@@ -37,6 +37,8 @@ function tide_dashboard_user_login($account) {
   $destination = \Drupal::service('redirect.destination')->get();
   if (!(strpos($destination, '/oauth/authorize') !== FALSE)) {
     $request = \Drupal::service('request_stack')->getCurrentRequest();
-    $request->query->set('destination', '/admin/workbench');
+    if (($request->query->has('destination')) === FALSE) {
+      $request->query->set('destination', '/admin/workbench');
+    }
   }
```
FleetAdmiralButter, line 43: Hey @edyuenyw,

Can we simplify the logic in this a bit?

Instead of checking for the existence of the destination /oauth/authorize, then checking the parameter again, we could just do:

```
function tide_dashboard_user_login($account) {
    if (($request->query->has('destination')) === FALSE) {
      $request->query->set('destination', '/admin/workbench');
    }
}
```

This will prevent a redirect to the workbench whenever the original request already has a different destination set.

FleetAdmiralButter (changes requested):

modules/tide_dashboard/tide_dashboard.module
edyuenyw: That looks cleaner. Will start making the change.

FleetAdmiralButter (approved):

PR #300, opened 2022-02-25
[SDPA-5452] CMS review page pagination

mayurngondhkar:
** Jira ** - https://digital-engagement.atlassian.net/browse/SDPA-5452

Changes:

Change the Diff modules revision_pager_limit to 16 in settings

image

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -1379,8 +1379,7 @@ function tide_core_update_8058() {
   if (\Drupal::moduleHandler()->moduleExists('diff')) {
     $config = \Drupal::configFactory()
       ->getEditable('diff.settings');
-    $value = $config->get('general_settings.revision_pager_limit');
-    if ($value) {
+    if (!$config->isNew() && !empty($config->get('general_settings.revision_pager_limit'))) {
       $config->set('general_settings.revision_pager_limit', 16)->save();
     }
   }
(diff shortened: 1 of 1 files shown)


tide_core.install
```diff
@@ -1371,3 +1371,17 @@ function tide_core_update_8057() {
...
 }
+
+/**
+ * Changes the diff modules general_settings.revision_pager_limit to 16.
+ */
+function tide_core_update_8058() {
+  if (\Drupal::moduleHandler()->moduleExists('diff')) {
+    $config = \Drupal::configFactory()
+      ->getEditable('diff.settings');
+    $value = $config->get('general_settings.revision_pager_limit');
```
anthony-malkoun, line 1382: If this value is not set, does this mean `$value` will be falsey? Or are we only setting it if it has some other value?

anthony-malkoun (commented):
I think you might want to check that `$config` exists rather than if `$value` does.

mayurngondhkar:
> I think you might want to check that `$config` exists rather than if `$value` does.

Added the check.

Stumbled across this resource  https://www.computerminds.co.uk/drupal-code/loading-config-programmatically

tide_core.install
mayurngondhkar: Idea was to check if $value was not empty.

anthony-malkoun (approved):
Looking good. Approved. Thanks for explaining the default being 50 and the link to https://www.computerminds.co.uk/drupal-code/loading-config-programmatically

PR #304, opened 2022-03-09
Remove `login_redirect_per_role` module from the install hook

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SRM-326

### Change
remove `tide_core_update_8048` hook so that it wouldn't be installed during the installation of `tide_core`.

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -1190,6 +1190,13 @@ function tide_core_update_8047() {
 }
 
 /**
+ * Set default home page and login redirect.
+ */
+function tide_core_update_8048() {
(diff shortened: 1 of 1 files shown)


anthony-malkoun:
Do you have to leave the hook there and just make it empty?

vincent-gao:
> Do you have to leave the hook there and just make it empty?

the latest number is 8057, it should be fine just left it empty. but yeah, I will leave the hook there, safer.

FleetAdmiralButter (approved):

PR #306, opened 2022-04-13
Deprecated _tide_import_single_config method

vincent-gao:
### Motivation
This function breaks the UUID sometimes, it should be removed before tide_core 4.x


anthony-malkoun (approved):
😍

sonnykt:
agree that the module should be deprecated. But for the UUID issue, this function should be used in conjunction with `_tide_retrieve_config_uuid` to preserve the UUID.

PR #309, opened 2022-04-28
[SDPA-6028] Added tide auto logout module.

MdNadimHossain:
### JIRA
https://digital-engagement.atlassian.net/browse/SDPA-6028

### Changes
Added autologout module in composer json

### Related PR
Using baywatch to enable this module and set default timeout value for all SDP projects.
https://github.com/dpc-sdp/baywatch/pull/58


sonnykt:
I wonder if it's necessary to create a Tide module just to enable a contrib module.

MdNadimHossain:
> I wonder if it's necessary to create a Tide module just to enable a contrib module. Should we use baywatch to enable the module and set the appropriate setting?

Hey @sonnykt 
My thinking was, if we have to put any custom logic in the future or any custom settings, it would be better to separate that. But if there is no value I will just enable it using baywatch. I wasn't sure that is why didn't put for review yet. But thanks for reviewing it anyway. :)

anthony-malkoun:
> I wonder if it's necessary to create a Tide module just to enable a contrib module. Should we use baywatch to enable the module and set the appropriate setting?

Why not just enable it in `tide_core`? Or if this is a baywatch/SDP only thing then should the composer require be there too?

sonnykt:
> > I wonder if it's necessary to create a Tide module just to enable a contrib module. Should we use baywatch to enable the module and set the appropriate setting?
> 
> Why not just enable it in `tide_core`? Or if this is a baywatch/SDP only thing then should the composer require be there too?

If it's in tide_core, *all* SDP sites will automatically have the module enabled. Is it a crucial module for all sites?

anthony-malkoun:
The point is, it's either in baywatch _or_ tide_core. All _SDP_ sites are getting it either way. Do all _Tide_ sites get it, is the question. And is this a different group in reality?

sonnykt:
> The point is, it's either in baywatch _or_ tide_core. All _SDP_ sites are getting it either way. Do all _Tide_ sites get it, is the question. And is this a different group in reality?

Are Health sites managed with the same Baywatch policies as other SDP sites?

Anyway, adding the modules as a dependency of tide_core makes it impossible for some sites to disable it if they wish. If I recall correctly, tide_site_restriction used to be a dependency and fully independent sites with just a single site term were not able to disable the module.

anthony-malkoun:
> Are Health sites managed with the same Baywatch policies as other SDP sites?

Yep. They're on SDP infrastructure so they get Baywatch.

MdNadimHossain:
I think enabling it in baywatch for all SDP sites will be good enough. The module sits on tide_core composer json anyway. so if any other tide sites outside of SDP needs it, they can just enable it separately according to their needs.

MdNadimHossain:
Moving this to baywatch as this module considered only for SDP sites

PR #311, opened 2022-05-09
[SDPA-5945] Updated iframe window for CKEditor.

MdNadimHossain:
### JIRA: 
https://digital-engagement.atlassian.net/browse/SDPA-5945

### Changes:
1) Added "title" in the text formats allowed html section for iframe.
2) Added js to change the "Advisory title" to Title
3) Also js is used to put the title at the top of the iframe window

### Screenshot
Screen Shot 2022-05-09 at 1 45 34 pm


js/node_iframe.js
```diff
@@ -0,0 +1,40 @@
...
+  Drupal.behaviors.nodeIFrame = {
+    attach: function (context, settings) {
+      if (!window.CKEDITOR) {
+        return;
+      }
+      CKEDITOR.on( 'dialogDefinition', function( ev ) {
+        // Take the dialog name and its definition from the event data.
+        var dialogName = ev.data.name;
+        var dialogDefinition = ev.data.definition;
+        console.log(dialogName);
```
vincent-gao, line 27: is this needed?

MdNadimHossain: lol nope, let me clean up

tide_core.install
```diff
@@ -1384,3 +1384,27 @@ function tide_core_update_8058() {
...
+    'filter.format.summary_text',
+  ];
+  foreach ($filter_ids as $filter_id) {
+    $filter = $config_factory->getEditable($filter_id);
+    if ($filter->isNew()) {
+      continue;
+    }
+
+    $value = $filter->get('filters.filter_html.settings.allowed_html');
+    $replaced = str_replace('<iframe frameborder height scrolling src width>', '<iframe frameborder height scrolling src width title>', $value);
```
anthony-malkoun, line 1406: Is this going to be brittle if anyone has changed the attributes allowed for an iFrame? Should it be a regex on the value that captures `<iframe (.*)>` and updates to `<iframe $1 title>`?

MdNadimHossain: it won't break but it will hide whatever is not allowed. It just adds a `display:none` in style to hide that field. Also if we put the wild card in, it will allow all the iframe fields to show up, which the content team does not want to have. For example, this is how the iframe window looks with all the fields 
Screen Shot 2022-05-09 at 5 04 31 pm

edyuenyw (approved):
LGTM

PR #314, opened 2022-05-17
[SDPA-6050] Fix stat grid default 2 blocks not showing

mayurngondhkar:
**Jira** - https://digital-engagement.atlassian.net/browse/SDPA-6050

Changes
1) Added patch in tide_core to provide the ability to add default paragraph count setting and logic to print out paragraph forms in a loop.


anthony-malkoun (approved):
LGTM. Please do try and progress the issue that this patch is against. It would be good if it was accepted into the modules main code.

PR #320, opened 2022-05-25
[SDP2-77] tide_jira updates

FleetAdmiralButter:
### JIRA
https://digital-engagement.atlassian.net/browse/SDPSUP-4386

1) Update install config to point to the new Jira instance
2) Implement new Jira field mappings for site, sub-site, departments and content types
3) Misc formatting improvements

### DIFF
--- a/modules/tide_jira/src/Form/TideJiraConfigForm.php
+++ b/modules/tide_jira/src/Form/TideJiraConfigForm.php
@@ -69,6 +69,47 @@ class TideJiraConfigForm extends ConfigFormBase {
       '#title' => $this->t('Issue Type'),
     ];
 
+    $form['field_mappings'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('Jira field mappings'),
+    ];
+
+    $form['field_mappings']['content_type'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('content_type'),
+      '#title' => $this->t('Content Type Field ID'),
+    ];
+
+    $form['field_mappings']['node_id'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('node_id'),
+      '#title' => $this->t('Node ID Field ID'),
+    ];
+
+    $form['field_mappings']['site'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('site'),
+      '#title' => $this->t('Site Field ID'),
+    ];
+
+    $form['field_mappings']['site_section'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('site_section'),
+      '#title' => $this->t('Site Section Field ID'),
+    ];
+
+    $form['field_mappings']['page_department'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('page_department'),
+      '#title' => $this->t('Page Department Field ID'),
+    ];
+
+    $form['field_mappings']['editor_department'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('editor_department'),
+      '#title' => $this->t('Editor Department Field ID'),
+    ];
+
     $form['no_account_email'] = [
       '#type' => 'fieldset',
       '#title' => $this->t('Email to use when user does not have an active JIRA account.'),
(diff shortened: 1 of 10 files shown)


modules/tide_jira/src/Form/TideJiraConfigForm.php
```diff
@@ -69,6 +69,47 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+
+    $form['field_mappings']['site_section'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('site_section'),
+      '#title' => $this->t('Site Section Field ID'),
+    ];
+
+    $form['field_mappings']['page_department'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('page_department'),
```
MdNadimHossain, line 103: @FleetAdmiralButter what happens if it can not get the department value, then what get sets into the default value. e.g department field is only on content-vic and what if other projects using the tide_jira and doesn't have the department field.

modules/tide_jira/src/Form/TideJiraConfigForm.php
```diff
@@ -69,6 +69,47 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#title' => $this->t('Page Department Field ID'),
+    ];
+
+    $form['field_mappings']['editor_department'] = [
+      '#type' => 'textfield',
+      '#default_value' => $config->get('editor_department'),
```
MdNadimHossain, line 109: same question for this default value as well!

modules/tide_jira/src/TideJiraAPI.php
```diff
@@ -216,9 +235,59 @@ private function getRevisionInfo(NodeInterface $node) {
       'is_new' => $node->isNew() ? 'New page' : 'Content update',
       'notes' => $node->getRevisionLogMessage(),
       'preview_links' => $this->getPreviewLinks($node, TRUE),
+      'site' => $this->entityTypeManager->getStorage('taxonomy_term')->load($node->get('field_node_primary_site')->getValue()[0]['target_id'])->getName(),
+      'site_section' => $this->getSiteSections($node),
+      'page_department' => $this->entityTypeManager->getStorage('taxonomy_term')->load($node->get('field_department_agency')->first()->getValue()['target_id'])->getName(),
```
MdNadimHossain, line 240: Does it need a check first if this exist and then get the value !

modules/tide_jira/src/TideJiraAPI.php
```diff
@@ -275,29 +342,25 @@ private function getAuthorInfo(NodeInterface $node) {
...
 
-Department:   $department
-
-Page information
+*Page information*
 
 Page name:     $title
 
-CMS URL:         https://content.vic.gov.au/node/$id
+CMS URL:         https://content.vic.gov.au/node/$id/revisions
```
MdNadimHossain, line 359: is there a way to get the CMS URL automated way, so that if it is not content-vic and some other project, it can get that CMS URL as well?

MdNadimHossain (commented):
@FleetAdmiralButter just few informative questions. Trying to understand few things :)

modules/tide_jira/src/Form/TideJiraConfigForm.php
FleetAdmiralButter: Hey @MdNadimHossain, this is the config form for mapping custom fields in Jira - it's not looking at the fields in Drupal. 

It's setting the page_department and editor_department values defined here: https://github.com/dpc-sdp/tide_core/blob/feature/SDP2-77-tide-jira-migration/modules/tide_jira/config/install/tide_jira.settings.yml. These will be set as the default upon installation as the field IDs are shared across projects in a Jira instance.

modules/tide_jira/src/TideJiraAPI.php
FleetAdmiralButter: Hi @MdNadimHossain,

As discussed, these fields are required on content-vic and should have a value.

If they're not available (i.e if deployed on another project), we can't just pass an empty string as this would affect reporting.

I've added further detail to the README to detail setup of these required fields.

modules/tide_jira/src/TideJiraAPI.php
FleetAdmiralButter: Fixed!

modules/tide_jira/src/TideJiraAPI.php
```diff
@@ -271,33 +343,31 @@ private function getAuthorInfo(NodeInterface $node) {
...
 
-Department:   $department
-
-Page information
+*Page information*
 
 Page name:     $title
 
-CMS URL:         https://content.vic.gov.au/node/$id
+CMS URL:         $host/node/$id/revisions
```
MdNadimHossain, line 366: :+1:

MdNadimHossain (approved):
LGTM now

PR #323, opened 2022-06-10
[SDPA-5959] Move CMS help block install to tide profile

christopher-hopper:
Ticket: https://digital-engagement.atlassian.net/browse/SDPA-5959

## Changes

- Remove install CMS Help block from `tide_cms_support`

## Motivation

During fresh install of `tide_core` and `tide_cms_support` the Admin theme is not available, so installation of the CMS Help block fails. 

See error during builds -
Screen Shot 2022-06-10 at 9 33 18 am

Moving this block placement into the Tide Profile install process where the Admin theme will be available.


MdNadimHossain (approved):
Good one, thanks.

PR #325, opened 2022-06-15
[SDPA-6303] Fixes an issue that causes link field broken

vincent-gao:
### Jira
https://digital-engagement.atlassian.net/browse/SDPA-6303

### Issue
see the ticket.

### DIFF
--- a/src/EntityAutocompleteMatcher.php
+++ b/src/EntityAutocompleteMatcher.php
@@ -4,7 +4,6 @@ namespace Drupal\tide_core;
 
 use Drupal\Component\Utility\Html;
 use Drupal\Component\Utility\Tags;
-use Drupal\Core\Entity\Plugin\EntityReferenceSelection\Broken;
 use Drupal\Core\Entity\EntityAutocompleteMatcher as CoreEntityAutocompleteMatcher;
 
 /**
@@ -18,7 +17,7 @@ class EntityAutocompleteMatcher extends CoreEntityAutocompleteMatcher {
   public function getMatches($target_type, $selection_handler, $selection_settings, $string = '') {
     // Match the selection handler for either 'default' or 'default:node'.
     if ($target_type == 'node' && $selection_handler == 'default') {
-      $selection_handler = 'tide_core';
+      $selection_handler = 'tide:node';
     }
     $matches = [];
     $options = $selection_settings + [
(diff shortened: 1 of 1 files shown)


mayurngondhkar (approved):
LGTM

src/EntityAutocompleteMatcher.php
```diff
@@ -27,6 +28,12 @@ public function getMatches($target_type, $selection_handler, $selection_settings
 
     $handler = $this->selectionManager->getInstance($options);
 
+    // If the handler cannot be found, returns the default one.
+    if ($handler instanceof Broken) {
```
christopher-hopper, line 32: @vincent-gao 
Just wondering under what circumstances is Drupal not able to load the actual EntityReferenceSelection plugin? Are we covering up some other config or code problem? Which plugin was it trying to load when it returned "Broken" instead?

christopher-hopper: I read the bug report ticket but couldn't see any clues.

christopher-hopper: @vincent-gao 
I think I see why it cannot load the EntityReferenceSelection handler. There's no hander called "tide_core". 

It should be looking for "tide:node" not "tide_core" at line 20.

The handler is defined here:
https://github.com/dpc-sdp/tide_core/blob/develop/src/Plugin/EntityReferenceSelection/TideNodeSelection.php 

```php
/**
 * Provides labels for the node entity type.
 *
 * @EntityReferenceSelection(
 *   id = "tide:node",
 *   label = @Translation("Tide node selection"),
 *   entity_types = {"node"},
 *   group = "tide",
 *   weight = 1,
 *   base_plugin_label = "Tide"
 * )
 */
```

If you fix that, it should no longer return the "broken" handler as a fallback I think. Can you try that out?

christopher-hopper (changes requested):

christopher-hopper:
@vincent-gao 
I think this might be the fix:
https://github.com/dpc-sdp/tide_core/pull/326/files 

Haven't tested it, but I think this is what's going on.

src/EntityAutocompleteMatcher.php
vincent-gao: Thanks @christopher-hopper ! I think it works.

christopher-hopper (approved):

PR #326, opened 2022-06-15
SDPA-6303: Fix broken handler plugin id

christopher-hopper:
Suggested change to PR dpc-sdp/tide_core#325


vincent-gao (approved):
Thanks @christopher-hopper , good change!

PR #333, opened 2022-08-15
[SDPAP-6365] update table of contents and display headings

MdNadimHossain:
### JIRA
https://digital-vic.atlassian.net/browse/SDPAP-6287

### Changes
1. Updated field_node_display_headings and field_show_table_of_content configs install.
2. Added a drupal database query to clean up the redundant values for the Display heading radio buttons.

### Related PRs
The independent modules that are using this fields, have updated the fields in their config/install directory.
https://github.com/dpc-sdp/tide_landing_page/pull/181
https://github.com/dpc-sdp/tide_publication/pull/22
https://github.com/dpc-sdp/tide_page/pull/28

### Test Link
https://nginx-php.pr-1429.content-vic.sdp4.sdp.vic.gov.au/


tide_core.install
```diff
@@ -1446,3 +1446,60 @@ function tide_core_update_8061() {
...
+  ];
+  foreach ($display_headings_tables as $table) {
+    $query = \Drupal::database();
+    if ($query->schema()->tableExists($table)) {
+      $query = $query->delete($table);
+      $orGroup = $query->orConditionGroup()
+        ->condition('field_node_display_headings_value', 'showH2AndAccordionTitle')
+        ->condition('field_node_display_headings_value', 'showH2AndH3AndAccordionTitle');
+      $query->condition($orGroup);
+      $query->execute();
```
MdNadimHossain, line 1467: As these fields already has value in it and they want to remove these two options that was added in the last release. The best way to avoid field migration and also not change the existing values for the first two radio buttons is just to clean up the redundant values from db if table exists. Then do the storage config updates for the field.

tide_core.install
```diff
@@ -1446,3 +1446,60 @@ function tide_core_update_8061() {
...
+    'field_node_display_headings' => [
+      'label' => t('Choose headings to display'),
+      'description' => '',
+    ],
+  ];
+  foreach ($field_configs as $field_config => $config) {
+    $field_storage = FieldStorageConfig::loadByName('node', $field_config);
+    if (!empty($field_storage)) {
+      $bundles = $field_storage->getBundles();
+      foreach ($bundles as $bundle) {
```
MdNadimHossain, line 1495: Although the field config seats in every module separately, but to apply the same changes on every module that using this field. It makes more sense to put this in tide_core, which will apply the changes to the content module that are project specific as well.

edyuenyw (approved):
LGTM

PR #348, opened 2022-10-25
SDPAP-5303: Navigating long menus in the CMS.

henrytranvan:
Jirra: https://digital-vic.atlassian.net/browse/SDPAP-5303

**Navigating long menus in the CMS**

As a CMS user I want to easily navigate through long drop-down menus in the CMS admin console to find the item I am looking for.

Background
As the CMS is growing the menu drop-downs are becoming bigger and bigger. There is currently no way of navigating through long navigation lists as items get cut off at the bottom of the screen. 

This is only an issue on two menu items:

Structure > Menus

Structure > Taxonomy

After a UX review [https://digital-vic.atlassian.net/browse/SDPA-5543 - Can't find link](https://digital-vic.atlassian.net/browse/SDPA-5543) the current solution is to:

remove the long secondary dropdowns (level 3) as they are not scalable, and

make users access the child options by clicking on the parent menu item (Menu and Taxonomy). 

The long term solution is to make these listing pages more usable via search and filter functions.

Assumptions
The chevron icons to the right of both Menus and Taxonomy are to be removed

The background colour of both Menus and Taxonomy will match the background colour of the other navigation options at the same level. 

Acceptance criteria
As a CMS approver in the CMS I can:

hover over the Structure main navigation item

click on Menus which takes me to the menus listing page

click on Taxonomy which takes me to the taxonomies listing page

I do not have any additional navigation drop-downs appear if I hover over Menus or Taxonomy

I cannot see chevron icons to the right of either Menus or Taxonomy

the background colour of Menus and Taxonomy navigation items is same colour as the other items at the same level (with no sub-items).

Solution direction
Scenario 1
[https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Menu!MenuLinkTree.php/function/MenuLinkTree%3A%3Aload/8.2.x](https://api.drupal.org/api/drupal/core/!lib/!Drupal/!Core/!Menu/!MenuLinkTree.php/function/MenuLinkTree%3A%3Aload/8.2.x)

This will disable the menu tree for all items under.

Scenario 2
Implement a hook that triggers when a menu item gets created.
If an item is under the menus it will not enable it and hence it will not appear in the dropdown.

### DIFF
--- a/src/Render/Element/AdminToolbar.php
+++ b/src/Render/Element/AdminToolbar.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Drupal\tide_core\Render\Element;
+
+use Drupal\Core\Menu\MenuTreeParameters;
+use Drupal\Core\Security\TrustedCallbackInterface;
+
+/**
+ * Define an admin toolbar.
+ *
+ * @package Drupal\tide_core\Render\Element
+ */
+class AdminToolbar implements TrustedCallbackInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function trustedCallbacks() {
+    return ['preRenderTray'];
+  }
+
+  /**
+   * Renders the toolbar's administration tray.
+   *
+   * This is a clone of admin_toolbar's
+   * admin_toolbar_prerender_toolbar_administration_tray() function, which uses
+   * setMaxDepth(5) instead of setMaxDepth(4).
+   *
+   * @param array $build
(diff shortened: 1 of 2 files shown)


anthony-malkoun:
@henrytranvan can you provide some detail in the PR description please? You can copy it from the JIRA ticket but I need to know what this change is doing, some testing steps and screenshots. Thanks!

PR #352, opened 2022-11-08
Feature SDPAP-5913 addition business fields

henrytranvan:
Jirra: https://digital-vic.atlassian.net/browse/SDPAP-5913

Addition business fields to CMS user account

As a content team member, when creating a CMS user profile for an external industry partner I would like to be able to add their business information so auditors can easily see who they work for and how they can contact them if needed. 

Problem
There is currently no ability to store additional business or change-management information within CMS user accounts.

Background
The content team is finding it difficult to manage user accounts from external agencies (industry partners and contractors) and require more information to be stored within their user account details. 

When looking up the user account of an industry partner, they would like to see and edit the following information:

Business name

Business contact number

Free-text multi-line notes field for an admin to record details of:

why they need an account

expiry details

internal contact name (person who engaged the contract)

project they are engaged for

other special conditions related to their account

Considerations
List any unanswered questions, considerations, concerns and risks.

 [SDPAP-5915: CMS user account expiry dateQA](https://digital-vic.atlassian.net/browse/SDPAP-5915) 

Acceptance criteria
As a content editor …

I can view my user account

I can edit my phone number

I can edit my business name

I cannot see the free text user notes field

As a site administrator or as an admin…

I can view a users account

I can edit the user's phone number

I can edit the user's business name

I can add notes into the free text field

I can edit existing notes in the free text field

Definition of Done
Acceptance criteria are met

Demo content is created/updated

Testing completed

Final functionality reviewed

All relevant user documentation is written

Print-friendly style tested

### DIFF
--- a/tide_core.module
+++ b/tide_core.module
@@ -22,6 +22,8 @@ use Drupal\views\ViewExecutable;
 use Drupal\workflows\Entity\Workflow;
 use Drupal\redirect\Entity\Redirect;
 use Drupal\Core\Cache\Cache;
+use Drupal\tide_core\Render\Element\AdminToolbar;
+use Drupal\Component\Utility\Html;
 
 /**
  * Implements hook_views_data().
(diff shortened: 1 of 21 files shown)


tide_core.module
```diff
@@ -152,6 +152,28 @@ function tide_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
       }
     }
   }
+  // Re-order the user_form field.
+  if ($form_id === 'user_form') {
```
henrytranvan, line 156: Hi @MdNadimHossain 
This is the new code for re-order the fields.
Thanks

tide_core.module
```diff
@@ -181,6 +203,25 @@ function tide_core_form_user_register_form_alter(&$form, FormStateInterface $for
   if (isset($form['account']['mail']['#required'])) {
     $form['account']['mail']['#required'] = TRUE;
   }
+  $form['account']['name']['#weight'] = 2;
```
henrytranvan, line 206: Hi @MdNadimHossain 
This is the new code for re-order the fields.
Thanks

tide_core.module
MdNadimHossain: why we are manipulating the weight in the alter hook, why can't we just make the changes in the config file directly using the update hook and then export the settings ?

tide_core.module
MdNadimHossain: same for this one - why we are manipulating the weight in the alter hook, why can't we just make the changes in the config file directly using the update hook and then export the settings ?

MdNadimHossain (changes requested):
@henrytranvan one question, why we are manipulating the weight in the alter hook, why not updating the weight using the update hook and then apply the changes using the config ?

tide_core.module
henrytranvan: Hi @MdNadimHossain 

The fields are 'based fields', defined by code. they can't configure the form display setting.
Screen Shot 2022-11-10 at 1 42 49 pm
The fields are grouped in side $form['account'] fieldset, custom field can't be config to show in the field set by UI.
The 'status', 'roles', 'notify', and 'password_policy_status' fields even doesn't show in the config form setting.
The 'notify' and 'password_policy_status' fields added by the contributor module, they not used on all sites so we need to use a code to check if the field exists.
That is why It needs code to set correctly order display.

Thanks

tide_core.module
henrytranvan: Hi @MdNadimHossain 

Good question, I think so too, but the fields are defined by code as 'base field', without the field setting UI. Please refer to the images below for more detail.
Screen Shot 2022-11-10 at 1 42 49 pm
The fields are grouped into a fieldset $form['account'], so custom fields can't add to the group by UI
'status', 'roles', 'notify', and  'password_policy_status' fields even doesn't show in the form settings
'notify', and  'password_policy_status' was added by contributor modules, they do not exist in all sites so we need code to check if the field is exist
That's why we need a custom code to set field orders correctly.
Thank you.
Henry

MdNadimHossain (approved):

PR #357, opened 2023-01-03
Updated simple oauth to latest version.

MdNadimHossain:


dev-tools.sh
```diff
@@ -34,5 +34,6 @@
 #
 # Uncomment and set the Dev-Tools's commit value and commit this change.
 # export GH_COMMIT=COMMIT_SHA
+export GH_COMMIT=ed88053154aaf46c43a5289b126621242d01afe8
```
edyuenyw, line 37: revert this?

edyuenyw (changes requested):

dev-tools.sh
MdNadimHossain: yup, once [this has been](https://github.com/dpc-sdp/dev-tools/pull/80) approved and released I will do that.

edyuenyw: Ah, this is the way.

edyuenyw (approved):

PR #359, opened 2023-01-09
[SDPAP-7116] Fixes an issue with approvers role

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-7116

### Issue
approvers don't have permission to break locked content.

### Change
adding `break content lock` permission to `approvers`

### DIFF
--- a/config/install/user.role.approver.yml
+++ b/config/install/user.role.approver.yml
@@ -29,6 +29,7 @@ permissions:
   - 'administer webform'
   - 'administer webform element access'
   - 'administer webform submission'
+  - 'break content lock'
   - 'create alert content'
   - 'create image_gallery content'
   - 'create landing_page content'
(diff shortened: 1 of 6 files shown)


vincent-gao:
@MdNadimHossain content-lock is enabled for all sites.  is that correct? I don't remember how content-lock gets enabled...

MdNadimHossain (approved):
I think it was enabled as a part of the tide_edit_protection module dependency by baywatch [here](https://github.com/dpc-sdp/baywatch/blob/d5c3c27824d77d4d6c1556ab0ac95f33b6612743/src/BaywatchOperation.php#L231)
This looks good to me. The permission might have been removed by the automation in the last release, we had similar kind of issues in the last release where it has removed few permissions.

krakerag (approved):
Also LGTM - however I don't have the same knowledge reference towards content lock that @MdNadimHossain does. From a code and config standpoint it looks fine.

PR #362, opened 2023-01-16
[SDPAP-6839] Add Dismissible Message Bar for CMS user

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-6400

### Problem/Motivation
There is currently no way for users to know if a CMS is undergoing maintenance or has upcoming maintenance or content freeze.

### Related PRs
https://github.com/dpc-sdp/content-vic-gov-au/pull/1455

### Screenshots
CleanShot 2023-02-01 at 10 17 10
CleanShot 2023-02-01 at 10 21 40

### TODO
The current suggested labels are hard coded, it may be worth providing a form for admin users in the near future...


MdNadimHossain:
wow it is already here, nice one @vincent-gao

modules/tide_site_alert/src/Plugin/Block/TideSiteAlertBlock.php
```diff
@@ -0,0 +1,45 @@
...
+    $build = parent::build();
+    foreach ($build as $key => $item) {
+      if (!is_numeric($key)) {
+        continue;
+      }
+      $value = NestedArray::getValue($build, ['#cache', 'tags', $key]);
+      if ($value && strpos($value, 'site_alert:') !== FALSE) {
+        preg_match('#[0-9]+$#', $value, $match);
+        $site_alert = SiteAlert::load(reset($match));
+        $build[$key]['#alert']['start'] = date('d/m/Y H:i:s', strtotime($site_alert->getStartTime()));
```
krakerag, line 32: I'm a bit of a stickler for DateTimeImmutable since you can trust the time you set on it hasn't changed.

modules/tide_site_alert/tide_site_alert.module
```diff
@@ -0,0 +1,70 @@
...
+      'template' => 'tide-site-alert',
+    ],
+  ];
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function tide_site_alert_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  if ($form_id === 'site_alert_add_form' || $form_id === 'site_alert_edit_form') {
```
krakerag, line 25: Usually better to refactory your if statements so you can return immediately and then have the code nested lower. It's a bit of an anti-pattern to have your entire function nested in an if statement

modules/tide_site_alert/tide_site_alert.module
```diff
@@ -0,0 +1,70 @@
...
+    $form['label']['widget'][0]['value']['#suffix'] = '</div>';
+  }
+}
+
+/**
+ * The callback for the form alter above.
+ */
+function tide_site_alert_textfield_callback(&$form, FormStateInterface $form_state) {
+  $selected = $form_state->getValue('suggested_labels');
+  if ($selected != 'Other') {
```
krakerag, line 63: I like to set something and then only have the if statement. Elses can normally be removed when it's just setting one value or another.

krakerag (approved):
LGTM @vincent-gao. A couple of code suggestions but it is fine the way it is.

vincent-gao:
> wow it is already here, nice one @vincent-gao

Thanks for the CR @MdNadimHossain

modules/tide_site_alert/src/Plugin/Block/TideSiteAlertBlock.php
vincent-gao: Hey @krakerag , thanks for the CR. 
It is always good to learn new things! 
I have made a further change trying to use `DateTimeImmutable` class. please take a look.

modules/tide_site_alert/tide_site_alert.module
vincent-gao: This reminds me that I should switch to using another hook!

modules/tide_site_alert/src/Plugin/Block/TideSiteAlertBlock.php
```diff
@@ -29,8 +29,11 @@ public function build() {
       if ($value && strpos($value, 'site_alert:') !== FALSE) {
         preg_match('#[0-9]+$#', $value, $match);
         $site_alert = SiteAlert::load(reset($match));
-        $build[$key]['#alert']['start'] = date('d/m/Y H:i:s', strtotime($site_alert->getStartTime()));
-        $build[$key]['#alert']['end'] = date('d/m/Y H:i:s', strtotime($site_alert->getEndTime()));
+        $dateTimeImmutable = new \DateTimeImmutable();
```
krakerag, line 32: Hmm I think this might not be correct.

DateTimeImmutable should be replacing date() - you're creating a new DateTimeImmutable object to work with, to create a formatted output for your two build variables. It should probably be more like this:

```
$build[$key]['#alert']['start'] = new \DateTimeImmutable($site_alert->getStartTime())->format('d/m/Y H:i:s');
$build[$key]['#alert']['end'] = new \DateTimeImmutable($site_alert->getEndTime())->format('d/m/Y H:i:s');
```

With your code, you're creating a new object based on time now(), and then modifying it by your start times/end times.

Does this make sense?

PR #366, opened 2023-01-24
Updated plugin implementation of entity_reference paragraphs widget.

MdNadimHossain:
### Jira
https://digital-vic.atlassian.net/browse/SRM-664

### Problem/Motivation
The media: buying for vic content type uses a link paragraph type which has a widget type set to PresetParagraphsWidget. tide_core has a custom plugin for this and [this place](https://github.com/dpc-sdp/tide_core/blob/develop/src/Plugin/Field/FieldWidget/PresetParagraphsWidget.php#L104) gets the preset_number from field config [here](https://github.com/dpc-sdp/content-vic-gov-au/blob/develop/config/sync/core.entity_form_display.node.media_buying_for_vic.default.yml#L338) and before it used to return a numeric value instead of empty string. With all the changes, its now returns exactly an empty string and which triggers the above error [in this line](https://github.com/dpc-sdp/tide_core/blob/develop/src/Plugin/Field/FieldWidget/PresetParagraphsWidget.php#L213)

### Fix
Change how sets up the #max_delta value, only do -1 when the $max value is a int else sets it as an empty string.

### DIFF
--- a/src/Plugin/Field/FieldWidget/PresetParagraphsWidget.php
+++ b/src/Plugin/Field/FieldWidget/PresetParagraphsWidget.php
@@ -210,7 +210,7 @@ class PresetParagraphsWidget extends ParagraphsWidget {
       '#required' => $this->fieldDefinition->isRequired(),
       '#field_name' => $field_name,
       '#cardinality' => $cardinality,
-      '#max_delta' => $max - 1,
+      '#max_delta' => is_int($max) ? ($max - 1) : '',
     ];
 
     if ($this->realItemCount > 0) {


src/Plugin/Field/FieldWidget/PresetParagraphsWidget.php
```diff
@@ -210,7 +210,7 @@ public function formMultipleElements(FieldItemListInterface $items, array &$form
       '#required' => $this->fieldDefinition->isRequired(),
       '#field_name' => $field_name,
       '#cardinality' => $cardinality,
-      '#max_delta' => $max - 1,
+      '#max_delta' => is_int($max) ? ($max - 1) : '',
```
krakerag, line 213: LGTM @MdNadimHossain

krakerag (approved):

src/Plugin/Field/FieldWidget/PresetParagraphsWidget.php
anthony-malkoun: Technically don't need the brackets around `$max - 1` but they're not doing any harm :)

anthony-malkoun (approved):
👍

src/Plugin/Field/FieldWidget/PresetParagraphsWidget.php
MdNadimHossain: yeah it is there for just visual purpose :p

PR #369, opened 2023-01-30
[SDPAP-7132] Automated config export cause drupal WSOD error.

edyuenyw:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-7132

### Problem/Motivation
There is a bug in an “Automated config export” script that added a Views config file containing a single line “status: false”. This config was imported by Drupal and resulted in an uncaught exception - White Screen of Death - error for every page load on the CMS.
When trying to change the "status", the config file could not be found, resulting in creation of an empty config file and inserted a single line "status: false".

### Fix
Added views.view.files to config/optional so during install, it will find the config. Config file has set "status: false".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,7 +69,7 @@
         "drupal/workbench": "^1.3",
         "drupal/wysiwyg_template": "^3.0",
         "drupal/dynamic_entity_reference": "^1.7",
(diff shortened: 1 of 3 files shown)


anthony-malkoun (approved):
Good find. This has been a pain for a while now.

PR #371, opened 2023-02-14
[SRM-685] Add custom callout CKEDITOR5 plugin.

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SRM-685

### Problem/Motivation
Drupal10 only supports CKEDITOR5, and the underlying architecture of CKEDITOR5 is entirely different from CKEDITOR4. 

### Fix
Add a custom module tide_ckeditor


modules/tide_ckeditor/js/ckeditor5_plugins/callout/src/calloutcommand.js
```diff
@@ -0,0 +1,17 @@
...
+        model.change(writer => {
+            const blocks = Array.from(document.selection.getSelectedBlocks())
+            for ( const block of blocks ) {
+                if ( !block.is( 'element', 'tidecallout' ) ) {
+                    writer.rename( block, 'tidecallout' );
+                }
+            }
+        });
+    }
+}
```
krakerag, line 17: Just a thought - some of these js files are 2 space indent and some are 4. Do we have a preferred style for JS files in a Drupal module? Is there a Drupal standard we should adopt here?

ping @anthony-malkoun

krakerag (commented):
Just one thought, otherwise it looks above board to me!

modules/tide_ckeditor/js/ckeditor5_plugins/callout/src/calloutcommand.js
anthony-malkoun: Hmm, I thought there was some `.editorconfig` for this. Worth looking into.

vincent-gao: hey @krakerag , @anthony-malkoun . just reformat the code to follow Drupal js style.

krakerag: I think that works. I might do some looking up as to what the Drupal standard is for in-project JS

krakerag (approved):
LGTM @vincent-gao

PR #383, opened 2023-02-27
Commented updated hook.

MdNadimHossain:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-7108

### Problem/Motivation
1. It causes issues for tide_site as tide_site not dependant on tide_landing_page  and creates a garbage field as the field is not provided by tide_core. This hook should have been added in tide_landing_page
2. It causes issues for tide_landing_page as well, cause tide_landing_page already has this field in the active config and then tide_core tries to add it again while running this hook and creates conflicts

### Fix
1. Commenting out the hook so that it does not run this on other modules where tide_core is added as a dependency.

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -1555,8 +1555,7 @@ function tide_core_update_8064() {
 }
 
 /**
- * tide_core_update_8065 is removed.
- * See ticket SDPAP-7108 for more details.
+ * See ticket SDPAP-7108, tide_core_update_8065 is removed.
  */
 
 /**
--- a/tide_core.install
+++ b/tide_core.install
@@ -1558,12 +1558,17 @@ function tide_core_update_8064() {
  * Prevent new entity reference creation.
  */
 function tide_core_update_8065() {
-  $config_factory = \Drupal::configFactory();
-  $config = $config_factory->getEditable('field.field.block_content.media_gallery.field_topic');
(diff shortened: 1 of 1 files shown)


MdNadimHossain:
I just learned that update hook runs on a fresh install for the modules that are added as a dependency. Also this issue is an example of why a related update hook should be added to the module by which the related fields are provided.

anthony-malkoun (changes requested):
Just remove it. Have a comment referring to the ticket if you want but remove the code.

MdNadimHossain:
> I just learned that update hook runs on a fresh install for the modules that are added as a dependency. Also this issue is an example of why a related update hook should be added to the module by which the related fields are provided.

I take this statement back, it does not runs the update hook during fresh install. tide_core has a logic set in [here](https://github.com/dpc-sdp/tide_core/blob/develop/tide_core.install#L89) which runs all the update hooks except for the ones defined in the array. But I still believe this particular update hook should not be here as this module is not the field provider, so removing the code from here.

vincent-gao (approved):

vincent-gao:
LGTM @MdNadimHossain

anthony-malkoun:
Maybe too late now, but I think the function signature needs to be there. Might break or not run later hooks on a clean install.

PR #395, opened 2023-03-28
[SDPAP-7631] DateTimeComputed tries to compute from an empty value

vincent-gao:
Fixes an issue where DateTimeComputed tries to compute from an empty value

### Jira
https://digital-vic.atlassian.net/browse/SDPAP-7631

### Reproduce

URL: https://nginx-php.pr-206.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au/node/add/event

Issue: After expanding the Feature Image toggle, and then clicking the ‘Select images' button, the modal for image selection is not displayed - the user is redirected to the Body Content tab

Expected: modal window to select image is displayed

### Drupal issue
https://www.drupal.org/project/drupal/issues/3052608

### Todo
we may apply/reroll more patches till it is finally fixed.


MdNadimHossain (approved):
lol I added that for content-legislation in the last release. But good that it is appearing for all other projects as well, now we know the error is consistent :D

PR #415, opened 2023-06-18
[DDS-1474] Fix bootstrap system corruption under high traffic

FleetAdmiralButter:
### Jira

https://digital-vic.atlassian.net/browse/DDS-1474

### Problem/Motivation

https://www.drupal.org/project/drupal/issues/3207813

Due to the way the Shield module behaves when used in combination with the Key module, corruption of the bootstrap system can occur as Shield attempts to use entityTypeManager too early in the request lifecycle before module files are loaded. This can cause the module_implements cache to be populated with missing data and as a result, hooks such as hook_entity_type_alter are no longer effective.

### Fix

Patch ModuleHandler to load all module files if invoked from an http_middleware service while the `module_implements` cache is not yet populated.


FleetAdmiralButter:
CI failure seems unrelated, looks like a new linting rule?

image

composer.json
```diff
@@ -283,9 +283,6 @@
             },
             "drupal/site_alert": {
                 "Add ability to dismiss alert - https://www.drupal.org/project/site_alert/issues/3156557#comment-14930684": "https://www.drupal.org/files/issues/2023-02-19/3156557-67.patch"
-            },
-            "drupal/ckeditor_templates": {
-                "CKEditor 5 support for Content Templates - https://www.drupal.org/project/ckeditor_templates/issues/3273358#comment-14963779": "./patches/ckeditor_templates/ckeditor_templates_4.patch"
```
FleetAdmiralButter, line 287: I couldn't get this patch to apply to content-vic. Pulled it out for now, will re-add before I merge.

anthony-malkoun (approved):

PR #417, opened 2023-06-21
[SDPAP-6777]add view and drupal export module

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-6777
### Problem/Motivation
As a site administrator or administrator, I would like to be able to download a user list from CMS so that use the data for reporting or communications purposes.
### Fix

- Moving view_data_export module and patch from content_vic to tide_core
- Adding user_created_by_field module to have the user field of who created a role.
- Create hook update to enable views_data_export, user_created_by_field modules.
- Import user list data.

### Related PRs
Created a pr in tide_site_restriction to get site restriction field. https://github.com/dpc-sdp/tide_site_restriction/pull/46
Created a pr in Content-vic to add extra user fields https://github.com/dpc-sdp/content-vic-gov-au/pull/1550

pr for testing in content-reference
https://github.com/dpc-sdp/content-reference-sdp-vic-gov-au/pull/296
Content-reference : https://nginx-php.pr-296.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au/


MdNadimHossain:
@yeniatencio I don't see any permission added, how it is going to be controlled ? Also isn't it suppose to be added somewhere from the menu ?

yeniatencio:
> @yeniatencio I don't see any permission added, how it is going to be controlled ? Also isn't it suppose to be added somewhere from the menu ?

Hey @MdNadimHossain , I have added permission for site admin to be able to access the report in people menu after discussing with Sarah that site admin doesn't have access to report menu.

tide_core.install
```diff
@@ -1934,3 +1934,32 @@ function tide_core_update_8091() {
...
+/**
+ * Import user list data.
+ */
+function tide_core_update_8094() {
+  $view = 'views.view.user_list_data_export';
+  $config_path = drupal_get_path('module', 'tide_core') . '/config/install';
+  $source = new FileStorage($config_path);
+  $config_storage = \Drupal::service('config.storage');
+  $config_storage->write($view, $source->read($view));
+}
```
krakerag, line 1965: I think because the code footprint is small here, this is fine, but for future PR's can we use the pattern of moving work into a class and out of the install hooks? Something like https://github.com/dpc-sdp/tide_search/blob/develop/tide_search.install#L113 for example :)

krakerag (approved):
Looks good @yeniatencio just one comment for thoughts

tide_core.install
yeniatencio: Will do. Thanks @krakerag .

PR #426, opened 2023-07-19
Bug/update hook function call on null fix

MdNadimHossain:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -77,7 +77,7 @@
         "drupal/field_permissions": "^1.2",
         "lesstif/php-jira-rest-client": "2.6.0",
         "drupal/jira_rest": "^4.0@beta",
-        "drupal/ckeditor5_embedded_content": "^1.0@alpha",
+        "drupal/ckeditor5_embedded_content": "1.0.0-alpha3",
         "respect/validation": "dev-master",
         "giggsey/libphonenumber-for-php": "^8.13",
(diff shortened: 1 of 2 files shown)


modules/tide_ckeditor/tide_ckeditor.install
```diff
@@ -21,6 +21,8 @@ function tide_ckeditor_update_9001() {
   $role_ids = ['editor', 'approver', 'site_admin', 'contributor'];
   foreach ($role_ids as $role_id) {
     $role = Role::load($role_id);
-    $role->grantPermission('use text format advanced')->save();
+    if ($role) {
+      $role->grantPermission('use text format advanced')->save();
```
MdNadimHossain, line 25: contributor role does not exist for all projects, that is why this needs a check before assigning the permission, so that it does not throw null exception.

composer.json
```diff
@@ -77,7 +77,7 @@
         "drupal/field_permissions": "^1.2",
         "lesstif/php-jira-rest-client": "2.6.0",
         "drupal/jira_rest": "^4.0@beta",
-        "drupal/ckeditor5_embedded_content": "^1.0@alpha",
+        "drupal/ckeditor5_embedded_content": "1.0.0-alpha3",
```
MdNadimHossain, line 80: ck5 embed has a custom local patch for all SDP projects, that is why this version needs to be locked so that the patch gets applied

PR #428, opened 2023-07-27
[SDPAP-8027]update locale description in metatag

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-8027
### Problem/Motivation
Due to a deprecated drupal field, a hotfix was implemented to use the metatags Locale field to allow the FE to determine the correct font to display for the entered language.
### Fix
Add logic to a hook form alter to update locale field text in metatag.
### Related PRs
Testing pr  https://nginx-php.pr-305.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au/
### Screenshots

Screenshot 2023-08-02 at 11 58 16 am


MdNadimHossain (commented):
@yeniatencio why a local patch like that ? isn't there a way to use any hook to manipulate the label ? Adding patch like this will make it harder to maintain once the contributed module updates something in the same file. Have you explored the idea of using a hook for this purpose ?

MdNadimHossain:
This is just an example how we are manipulating the paragraph label here for that node view - https://github.com/dpc-sdp/tide_landing_page/blob/develop/tide_landing_page.module#L227

krakerag:
I agree with @MdNadimHossain in that having a patch makes it harder to maintain (nearly every other way apart from a patch is easier to maintain) but your example isn't really a great one Nadim due to it being a hook modification on a set of fields that we are in full control of in tide, so we are adding the field in the same contrib module as handling the hook to adjust the label. This means no other actor is in control of the field so using a hook to manage it is more graceful and maintainable.

@yeniatencio if you can find a way to use a hook rather than a patch, it would be good.

config/optional/metatag.metatag_defaults.node.yml
```diff
@@ -8,3 +8,4 @@ tags:
   description: '[node:summary]'
   canonical_url: '[node:url]'
   og_image: '[node:field_featured_image:entity:field_media_image:entity:url]'
+  og_locale: 'en-au'
```
krakerag, line 11: I actually don't think we use en-au anywhere.

This page here - https://digital-vic.atlassian.net/servicedesk/customer/article/2272657746 - lists the documentation that the content team use for managing this field and the codes they use. A good one might be 'en' for an example - but we don't need to set a default at all here.

The default for our english content should be an empty field.

krakerag (commented):
One thought on the defaults and on a hook based approach @yeniatencio

config/optional/metatag.metatag_defaults.node.yml
yeniatencio: Hi @krakerag and @MdNadimHossain , I have updated my pr for your reviews. Ta

MdNadimHossain (approved):

PR #429, opened 2023-08-01
[SDPAP-8093] remove deprecated modules before d10 update

MdNadimHossain:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-8093

### Problem/Motivation
Some of the module was deprecated in Drupal 9.4 and the code package reference is removed from Drupal 10. So those modules needs to be uninstalled first from SDP projects before installing the Drupal 10

### Fix
1. Removed Deprecated modules in 9.4 and added update hook to clean it from all existing projects.
2.

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -1921,7 +1921,7 @@ function tide_core_update_8090() {
 function tide_core_update_8091() {
   $modules = ['ckeditor', 'wysiwyg_template_core', 'color', 'hal', 'tide_spell_checker'];
   foreach ($modules as $module) {
-    // Check if queue_mail is both installed and enabled.
+    // Check if module is both installed and enabled.
     if (\Drupal::moduleHandler()->moduleExists($module)) {
       // If exists, uninstall module.
       \Drupal::service('module_installer')->uninstall([$module]);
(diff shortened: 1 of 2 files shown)


tide_core.install
```diff
@@ -1914,3 +1914,23 @@ function tide_core_update_8090() {
...
+/**
+ * Remove deprecated modules before d10 upgrade.
+ */
+function tide_core_update_8091() {
+  $modules = [
+    'ckeditor',
+    'wysiwyg_template_core',
+    'color',
+    'hal',
+    'tide_spell_checker',
```
MdNadimHossain, line 1927: `tide_spell_checker` Will not be supported anymore and the code will be cleaned up in the Drupal 10 upgrade

krakerag:
Do we need to remove them out of composer.json also for dependency checking?

MdNadimHossain:
> Do we need to remove them out of composer.json also for dependency checking?

It is removed in the core composer of d10. That is why we need to uninstall it first before updating the projects to d10. Else it will complain the package does not exist. So all these are marked as deprecated from 9.4, so first step will be uninstalling them from the cms while the package is still there. In d10 they are removed, so it will get cleaned up from the composer during the d10 upgrade

krakerag (approved):

anthony-malkoun:
> Do we need to remove them out of composer.json also for dependency checking?

The only problem with that is it has to be a two step process. The uninstall requires the code to be there.

PR #431, opened 2023-09-07
[SDPAP-8299] Added null check for moderation state.

MdNadimHossain:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-8299
### Problem/Motivation
Screenshot 2023-09-07 at 4 22 57 pm

### Fix
Null check for moderation state to avoid this kind behaviour for content types that does not have any moderation state.


anthony-malkoun (approved):
This seems like a no brainer to me. Approved.

PR #434, opened 2023-09-25
Updated deprectaed functions.

MdNadimHossain:
### Problem/Motivation
It is causing issues envs going down whenever it tries to use this. It was caught during a test in the reference site BE.


anthony-malkoun (approved):
👍

anthony-malkoun:
Assuming you did a global search for these @MdNadimHossain?

PR #437, opened 2023-10-02
Fix version checking block on helper method

krakerag:
We don't use D8 anymore, so there's a large block of code here not required for our purposes.

I noticed this while checking to see if there is a helper method to assist in config checking prior to import.


MdNadimHossain (approved):
good one 👍

PR #438, opened 2023-10-05
Update requirements to remove broken patch

krakerag:
A build over in tide_search (https://github.com/dpc-sdp/tide_search/pull/58) has discovered an issue with the patch here, so it should be removed since it's now in release.


vincent-gao (approved):
👍

MdNadimHossain (approved):

PR #440, opened 2023-10-10
[SDPAP-7910] MFA implementation v1.2

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SD-33

### Problem/Motivation
As an SDP administrator

I want multi-factor authentication to be enforced for all CMS users

So that SDP can continue to be at the forefront of security best practice
### Fix
The following changes have been implemented:
- Added TFA and Real AES modules
- Added TFA Drupal patches for cleaning up authenticator links
- Added config files on install and optional folders
- Added function in TideCoreOperation for enabling TFA necessary modules, and update settings.
- Added function in TideCoreOperation for disabling site alert on TFA routes.
- Fixed typo in tide_core.install
- Added hook_update for TFA setup.
- Added tfa email plugin.
- Override tfa email plugin methods to achieve some label changes.

**Note: this PR has been branch out from [feature/SRM-870-d10-update](https://github.com/dpc-sdp/tide_core/tree/feature/SRM-870-d10-update).**

### Related PRs
Testing PR https://github.com/dpc-sdp/content-reference-sdp-vic-gov-au/pull/330
Testing PR Link: https://nginx-php.pr-330.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au/

### Screenshots
Screenshot 2023-09-15 at 11 18 47 am

Screenshot 2023-09-15 at 11 16 34 am

Screenshot 2023-09-15 at 11 17 28 am

### TODO

- Currently, TFA works fine when using a file for the encryption key. However, when using an environment variable, generating the random key on base64 (as per documentation on the module), it throws a validation error due to size not matching. Furthermore, when fixing the key to match the size, Real AES module throws an Encrypt exception message: Integrity check failed. In summary, there's an error when using key type Encryption and Environment variable Base64 as a key provider.


config/install/key.key.tfa_encryption_key.yml
```diff
@@ -0,0 +1,16 @@
+langcode: en
+status: true
+dependencies: {  }
+id: tfa_encryption_key
+label: 'TFA encryption key'
+description: ''
+key_type: encryption
+key_type_settings:
+  key_size: 256
+key_provider: file
```
anthony-malkoun, line 10: Shouldn't the `key_provider` be `environment`?

tide_core.install
```diff
@@ -82,3 +85,24 @@ function tide_core_update_8097() {
     $editor->save();
   }
 }
+
+/**
+ * Setup TFA.
+ */
+function tide_core_update_8098() {
```
anthony-malkoun, line 92: @vincent-gao were we renumbering updates? Should this be `tide_core_update_10001()`?

config/install/key.key.tfa_encryption_key.yml
yeniatencio: Hi @anthony-malkoun , yes we had a conversation about it https://digital-engagement.slack.com/archives/C013GKSAE30/p1694676758318889 I will ping devops guys about it.

yeniatencio: @anthony-malkoun this has been updated to use env variable instead.

tide_core.install
vincent-gao: hey @anthony-malkoun , yes.  ideally, we should keep the update hook number consistent with the main version of Drupal.

src/TideCoreOperation.php
```diff
@@ -233,4 +234,76 @@ public function chagneDiffSettings() {
...
+      'encrypt.profile.tfa_encryption',
+    ];
+
+    $config_files_optional = [
+      'encrypt.settings',
+      'tfa.settings',
+    ];
+
+    foreach ($configs_files_install as $install) {
+      _tide_ensure_config($install, $config_install);
```
vincent-gao, line 284: This may not be related to this PR, but I think using `_tide_ensure_config` breaks the uuid for configurations. We should avoid abruptly overwriting its upstream configurations.

cc @anthony-malkoun / @krakerag

krakerag: I haven't used _tide_ensure_config but it does seem to forcibly reinstall a config, which would generate a fresh uuid every time.

It's probably great for brand new configuration, but if you're updating existing I think we have the paradigm in a lot of our libraries about using _tide_read_config and checking for load and updating that way?

vincent-gao: _tide_ensure_config uses `\Drupal::service('config.storage')->write($id, $config_data);` This type of writing cannot generate a new UUID. It simply writes the YAML file to the database.

> It's probably great for brand new configuration, but if you're updating existing I think we have the paradigm in a lot of our libraries about using _tide_read_config and checking for load and updating that way?
yes, @krakerag, we have to read the configurations to a variable and then let Drupal handle the rest.

and in this case, we should not define `encrypt.settings.yml` and `tfa.settings.yml` here since they are already defined by tide_core's upstream. @yeniatencio

tide_core.install
yeniatencio: Updated

src/TideCoreOperation.php
yeniatencio: @vincent-gao @krakerag  I have updated the code to use _tide_read_config for installing the encryption key and profile. 
Regarding `encrypt.settings.yml` and `tfa.settings.yml`, I still using _tide_ensure_config function as those files do not have a uuid. I need to defined those files and update them later on the hook update as the TFA module (when enable) will generate them with some basic values. It seems is not possible to add them on the array and use the same _tide_read_config as is not picking up simple configuration types.

vincent-gao: @yeniatencio This is not the correct approach. Removing default configurations and directly adding new ones breaks the change tracking mechanism.

vincent-gao (changes requested):
please see my comments.

src/TideCoreOperation.php
vincent-gao: hey @yeniatencio ,
I think you could modify the default configurations instead of overwriting them. In my opinion, this change should be a submodule of tide_core, and there should be a mechanism to `disable` or `turn off` the module, especially considering that this is a significant change for the SDP. This way, in case something goes wrong, you would have the option to disable it without breaking anything.
cc @krakerag

yeniatencio: @vincent-gao,
The default configuration gets saved in the db once the TFA module is enabled. are you saying to call a function to set the value of each field? isn't the same thing as overwriting them? Not really sure what you mean or if I'm misunderstanding how to do it. 
The TFA module already provides the drush commands to disable/enable the module: 

- Disable TFA with `drush config-set tfa.settings enabled 0`
- Enable TFA with `drush config-set tfa.settings enabled 1`

vincent-gao: 1. move all of your changes to a submodule, which can be named tide_tfa, under tide_core
2. remove `config/optional/encrypt.settings.yml`  and `config/optional/tfa.settings.yml`
3. move `config/install/encrypt.profile.tfa_encryption.yml` and `config/install/key.key.tfa_encryption_key.yml` to the `optional` directory.

Yes, they are different. Overwriting using `write` function will break `uuid` or `default_config_hash`, we should avoid this kind of operation.

yeniatencio: Hey @vincent-gao, I have updated my pr for your review. Thanks you for the explanation yesterday.

vincent-gao (changes requested):
hi @yeniatencio  please add tests.

tide_core.info.yml
```diff
@@ -73,6 +73,7 @@ dependencies:
   - simple_sitemap:simple_sitemap
   - tablefield:tablefield
   - tide_ckeditor:tide_ckeditor
+  - tide_tfa:tide_tfa
```
vincent-gao, line 76: tide_tfa shouldn't be here, which will complicate the dependencies.

vincent-gao (changes requested):

tide_core.info.yml
yeniatencio: removed

yeniatencio:
> hi @yeniatencio please add tests.

Hi @vincent-gao , yes I have added test and updated tide._core.info yml.

tests/behat/features/tide_2fa.feature
```diff
@@ -0,0 +1,34 @@
...
+    And I see the text "2FA Settings"
+    And I see field "edit-tfa-enabled"
+    And the "edit-tfa-enabled" checkbox should not be checked
+    And I see field "edit-tfa-forced"
+    And the "edit-tfa-forced" checkbox should be checked
+    Then I save screenshot
+    Examples:
+      | role               |
+      | administrator      |
+      | site_admin         |
```
vincent-gao, line 34: hey @yeniatencio 
please add a scenario where `tfa` is enabled and it blocks all of the users.

vincent-gao (changes requested):
Hey @yeniatencio ,
Please refer to my comments.

vincent-gao (approved):
Good work. Thanks ,@yeniatencio

anthony-malkoun (approved):
A pretty lightweight change considering how important it is. Looks good to me.

krakerag:
hi @MdNadimHossain can you remove me as a reviewer/subscriber on this PR :) Cheers!

PR #462, opened 2024-01-01
Fix issue presenting on login screen with class attributes

krakerag:
Fixes an issue when logging into CMS about the Request class attribute being added dynamically

image


vincent-gao (approved):
👍

PR #464, opened 2024-01-08
updated drupal core to 10.1.7

MdNadimHossain:
### Problem/Motivation
Couldn't update to 10.2.1 version as most of the patches we are using still not compatible with that version and it is not only a simple roll out for those patches. 10.1 has still coverage till June 2024 and all the patches works with it. Also the 10.2 has all the minor bug fixes, so we can do it later once all those patches are sorted. 

### Fix
Updated drupal core to 10.1.7


anthony-malkoun (approved):
I think this is the way forward. No rush to get to 10.2 at this stage.

krakerag:
Agree - 10.2 is still too edge I think.

Latest 10.1 is the go

krakerag (approved):

vincent-gao (approved):

PR #469, opened 2024-01-29
[SDPAP-8538] Fixes last update date issue

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SDPAP-8538

### Problem/Motivation
1. for this specific version of entity_embed, it doesn't support well for the wildcard, so let's just explicitly specify all tags we need. 
2. other issues are detailed in https://www.drupal.org/project/entity_embed/issues/3410132
3. it should be noted that this is only a temporary solution.

### Fix
https://www.drupal.org/project/entity_embed/issues/3410132#comment-15418003


yeniatencio (approved):
LGTM

PR #471, opened 2024-02-01
Locked respect/validation to latest release to avoid composer conflicts.

MdNadimHossain:
### Problem/Motivation
dev-master has an update for symfony/polyfill-mbstring which creates conflicts with drupal core 10.1.x dependency.
### Fix
Lock it to a stable latest release tag to avoid this conflicts.


mayurngondhkar (approved):

yeniatencio (approved):
Hi @MdNadimHossain , are we going to update it in tide_search?

MdNadimHossain:
> Hi @MdNadimHossain , are we going to update it in tide_search?

@yeniatencio do we need to update the tide_search ?

PR #478, opened 2024-03-06
fixed hook

yeniatencio:
### Problem/Motivation
Content-reference build was failing because hook update wrong configs files

### Fix
Update hook update
### Screenshots
Screenshot 2024-03-06 at 3 04 04 pm


MdNadimHossain (changes requested):
@yeniatencio the previous step was right, just need a slight modification, try like this - https://github.com/dpc-sdp/baywatch/blob/master/src/BaywatchOperation.php#L92
check first if the field exist, if it returns null then only createFromStorageRecord

MdNadimHossain (approved):

PR #481, opened 2024-04-11
Updated the patch to a new re-rolled version agaisnt the new release of entity embed.

MdNadimHossain:
### Problem/Motivation
The patch was outdated after the new version release of entity embed.
### Fix
Added the new re-rolled patch against 1.6 version of entity embed.


yeniatencio (approved):
LGTM.

PR #495, opened 2024-07-24
Removed the patch for user_expire as it has fixed

vincent-gao:
### Problem/Motivation
https://www.drupal.org/project/user_expire/issues/3253423

It has been fixed upstream, so we have removed it.


anthony-malkoun (approved):
How good is removing patches!

PR #498, opened 2024-07-29
Update token version

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SD-134
### Problem/Motivation
Patch doesn't apply in the last version of token as the code has been refactored. Therefore, downgrading to version 12 and update to latest available version of the patch.
### Fix

https://www.drupal.org/project/token/issues/2924873#comment-15705267
### Screenshots
Screenshot 2024-07-29 at 11 27 16 pm

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "drupal/scheduled_transitions": "^2.3",
         "drupal/link_field_autocomplete_filter": "^2.0",
         "drupal/queue_mail": "^1.5",
-        "drupal/token": "^1.11",
+        "drupal/token": "^1.15",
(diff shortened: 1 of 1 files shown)


vincent-gao (changes requested):
hey @yeniatencio 
Thanks for the changes.
I think we should not limit the version to a specific old version, especially when the latest version has passed the Drupal security tests. Please keep it at ^1.11 and update the patch.

vincent-gao (commented):
noice! @yeniatencio !

vincent-gao (approved):

PR #516, opened 2024-10-01
[SD-22]Updated hero image theme according header style

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SD-22
BE testing link: https://nginx-php.pr-1711.content-vic.sdp4.sdp.vic.gov.au/
### Problem/Motivation
We want to have these scenarios for customised header: 
1. Corner graphics - Light + Dark (Both options available and by default Light should be selected),
2. Full background image - Dark only (Only one option should be set) 
3. CTA - Light only (Only one option should be set).
4. Default appearance - Light only (Only one option should be set)

### Fix
1. Set field `field_landing_page_hero_theme` to mandatory.
2.  Remove a patch for drupal/jsonapi_extras as new released was done and patch is not needed anymore.
3. Fix typo in label : Here image theme.
4. Updated behat test.
5. Updated tide publication module to make it work in publication and publication page content type. 
### Screenshots
Screenshot 2024-10-03 at 10 01 49 am
Screenshot 2024-10-03 at 10 01 36 am
Screenshot 2024-10-03 at 10 01 25 am
Screenshot 2024-10-03 at 10 01 12 am


modules/tide_landing_page/js/hero_image_theme.js
```diff
@@ -0,0 +1,39 @@
...
+
+        if (defaultHeaderStyle === "fullwidth") {
+          imageTheme.value = "dark";
+        } else {
+          imageTheme.value = "light";
+        }
+      }
+    },
+  };
+})(Drupal);
```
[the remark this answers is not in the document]
yeniatencio: @MdNadimHossain agreed. I've already using conditional logic for enable/disable the `field_landing_page_hero_theme` as per the requirements. However, one of the requirements is to set the default value of one field based on the value of another field and the only [Drupal States](https://www.drupal.org/docs/drupal-apis/form-api/conditional-form-fields#statesproperty) that can be applied to a form field element are:
- enabled
- disabled
- required
- optional
- visible
- invisible
- checked
- unchecked
- expanded
- collapsed

As you can notice the default value can't be changed using states. Also, all the examples in landing page module is for changing one of the states in the list above but a default value.
Therefore, that's why I ended up writing JS code for it. Hope it makes sense now.

modules/tide_landing_page/js/hero_image_theme.js
```diff
@@ -0,0 +1,39 @@
+(function (Drupal) {
+  "use strict";
+
+  Drupal.behaviors.heroImageTheme = {
+    attach: function (context, settings) {
+      const imageTheme = document.querySelector(
```
MdNadimHossain, line 6: Use context to ensure that the behavior only attaches to elements in the current page context.

modules/tide_landing_page/js/hero_image_theme.js
```diff
@@ -0,0 +1,39 @@
...
+      );
+
+      if (imageTheme !== null) {
+        setImageTheme();
+
+        const headerStyles = document.querySelectorAll(
+          'input[data-drupal-selector^="edit-header-style-options"]'
+        );
+
+        if (headerStyles !== null) {
```
MdNadimHossain, line 17: The check if (headerStyles !== null) should instead check for headerStyles.length > 0. It’s better to check if any elements are found before adding listeners.

modules/tide_landing_page/js/hero_image_theme.js
```diff
@@ -0,0 +1,39 @@
...
+          headerStyles.forEach((style) => {
+            style.addEventListener("change", () => {
+              setImageTheme();
+            });
+          });
+        }
+      }
+
+      function setImageTheme() {
+        let defaultHeaderStyle = document.querySelector(
```
MdNadimHossain, line 27: Add a guard clause to prevent this error in case no radio buttons are checked

modules/tide_landing_page/js/hero_image_theme.js
```diff
@@ -0,0 +1,39 @@
...
+        setImageTheme();
+
+        const headerStyles = document.querySelectorAll(
+          'input[data-drupal-selector^="edit-header-style-options"]'
+        );
+
+        if (headerStyles !== null) {
+          headerStyles.forEach((style) => {
+            style.addEventListener("change", () => {
+              setImageTheme();
```
MdNadimHossain, line 20: Instead of repeatedly defining an anonymous function for style.addEventListener("change", ...), you can directly pass the reference to the setImageTheme function.

modules/tide_landing_page/js/hero_image_theme.js
```diff
@@ -0,0 +1,39 @@
...
+      const imageTheme = document.querySelector(
+        'select[data-drupal-selector^="edit-field-landing-page-hero-theme"]'
+      );
+
+      if (imageTheme !== null) {
+        setImageTheme();
+
+        const headerStyles = document.querySelectorAll(
+          'input[data-drupal-selector^="edit-header-style-options"]'
+        );
```
MdNadimHossain, line 15: The script accesses document multiple times. It’s a good practice to minimize global object access by caching elements that you reuse.

modules/tide_landing_page/js/hero_image_theme.js
```diff
@@ -0,0 +1,39 @@
...
+        if (headerStyles !== null) {
+          headerStyles.forEach((style) => {
+            style.addEventListener("change", () => {
+              setImageTheme();
+            });
+          });
+        }
+      }
+
+      function setImageTheme() {
```
MdNadimHossain, line 26: I would suggest moving the setImageTheme() function outside the main attach function for clarity and better separation of logic.

MdNadimHossain (changes requested):
@yeniatencio I have suggested some suggestions. Please have a look into the revised version and it should work in the same way - 
```
(function (Drupal) {
  "use strict";

  Drupal.behaviors.heroImageTheme = {
    attach: function (context, settings) {
      const imageTheme = context.querySelector(
        'select[data-drupal-selector^="edit-field-landing-page-hero-theme"]'
      );

      if (imageTheme !== null) {
        setImageTheme();

        const headerStyles = context.querySelectorAll(
          'input[data-drupal-selector^="edit-header-style-options"]'
        );

        if (headerStyles.length > 0) {
          headerStyles.forEach((style) => {
            style.addEventListener("change", setImageTheme);
          });
        }
      }

      function setImageTheme() {
        const selectedHeaderStyle = document.querySelector(
          'input[data-drupal-selector^="edit-header-style-options"]:checked'
        );

        if (!selectedHeaderStyle) {
          return; // Exit if no style option is selected.
        }

        let defaultHeaderStyle = selectedHeaderStyle.value;
        imageTheme.value = defaultHeaderStyle === "fullwidth" ? "dark" : "light";
      }
    },
  };
})(Drupal);
```

yeniatencio:
> @yeniatencio I have suggested some suggestions. Please have a look into the revised version and it should work in the same way -
> 
> ```
> (function (Drupal) {
>   "use strict";
> 
>   Drupal.behaviors.heroImageTheme = {
>     attach: function (context, settings) {
>       const imageTheme = context.querySelector(
>         'select[data-drupal-selector^="edit-field-landing-page-hero-theme"]'
>       );
> 
>       if (imageTheme !== null) {
>         setImageTheme();
> 
>         const headerStyles = context.querySelectorAll(
>           'input[data-drupal-selector^="edit-header-style-options"]'
>         );
> 
>         if (headerStyles.length > 0) {
>           headerStyles.forEach((style) => {
>             style.addEventListener("change", setImageTheme);
>           });
>         }
>       }
> 
>       function setImageTheme() {
>         const selectedHeaderStyle = document.querySelector(
>           'input[data-drupal-selector^="edit-header-style-options"]:checked'
>         );
> 
>         if (!selectedHeaderStyle) {
>           return; // Exit if no style option is selected.
>         }
> 
>         let defaultHeaderStyle = selectedHeaderStyle.value;
>         imageTheme.value = defaultHeaderStyle === "fullwidth" ? "dark" : "light";
>       }
>     },
>   };
> })(Drupal);
> ```

Hi @MdNadimHossain , Thanks for the suggestions. I have updated my pr as you requested.

MdNadimHossain (approved):

yeniatencio:
Hey @MdNadimHossain , I have updated my pr because there is a bug present when using states.  In our case it happens when a hero image is added, then the hero image theme dropdown was being disabled. 
For reference you can see here https://www.drupal.org/project/drupal/issues/1091852.
Now it works as requested: 

Screenshot 2024-10-28 at 4 22 54 pm

vincent-gao:
hey @yeniatencio 
I believe this can be achieved using [#ajax](https://www.drupal.org/docs/develop/drupal-apis/javascript-api/ajax-forms) without adding custom JavaScript code. As @MdNadimHossain pointed out, we should avoid managing additional JS solely for visual enhancements. You can also implement a submit handler to validate the form if needed.

vincent-gao:
for JS itself. 
1. Limit the DOM operations within the context. https://github.com/dpc-sdp/tide_core/pull/516/files#diff-74b152131cfda4b1eb7d53850fae87f6de6abcbc57cd5143247a4b8480acc46aR25
2. Use the `once` method to avoid repeatedly binding event listeners when behaviors are executed multiple times

vincent-gao (changes requested):
hey @yeniatencio 
Please review my comments and provide the test code for this new feature

PR #519, opened 2024-10-04
[SD-197] Enable reusable paragraph type

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-197
https://digital-vic.atlassian.net/browse/SD-332
https://digital-vic.atlassian.net/browse/SD-336
https://digital-vic.atlassian.net/browse/SD-370

### Changes
1. Enabled reusable paragraph module.
2. Added customised icon for reusable paragraph.
4. Altered paragraphs_library view to add site and department field.
5. Altered paragraphs_library_browser view to add site and department field.
6. Added 2 helper functions to update, import, and merge configurations from the module to active.


src/TideCoreOperation.php
```diff
@@ -258,4 +258,31 @@ public function enabledTideTfa() {
...
 
+  /**
+   * Enables paragraphs_library.
+   */
+  public function alterParagraphsLibrary() {
+    // Enabled paragraphs_library module.
+    if (!\Drupal::moduleHandler()->moduleExists('paragraphs_library')) {
+      /** @var \Drupal\Core\Extension\ModuleInstallerInterface $module_installer */
+      $module_installer = \Drupal::service('module_installer');
+      $module_installer->install(['paragraphs_library']);
```
MdNadimHossain, line 269: I think we should add a helper function to install and uninstall modules with proper try catch error handling. But we can do that separately, doesn't need to be a part of this.

MdNadimHossain (approved):
LGTM as most of them is just adding new configs and their related update hooks. Thanks for the helper function, we should make a practice to use them. Also I think we should add installing and uninstalling module helper function with proper error handling. The PR changes looks straightforward to me, as long as the testing passes, should be fine to merge. I am assuming this will be enabled for all projects, so making it part of a core dependency looks fine to me.

PR #520, opened 2024-10-10
Added code clean up and remove tfa enable with hook.

MdNadimHossain:
### Problem/Motivation
There is currently a conflict between the TFA password reset and the [PRLP](https://www.drupal.org/project/prlp) module. We need an SDP-specific patch to ensure the PRLP password reset functions correctly with TFA. Without this patch, the TFA module bypasses the PRLP service and instead calls Drupal core's default password reset service.

### Fix
1. We will stop auto-enabling the tide_tfa module for all projects. Although the TFA code works as expected, we will only enable it on a project-by-project basis after implementing the SDP-specific patch that resolves the password reset issue. This will ensure proper integration between TFA and PRLP.

2. Removed duplicate sate js addition in landing page module as that addition is moved to tide_core.module.

### DIFF
--- a/modules/tide_landing_page/tide_landing_page.module
+++ b/modules/tide_landing_page/tide_landing_page.module
@@ -109,11 +109,6 @@ function tide_landing_page_form_node_form_alter(&$form, FormStateInterface $form
       }
     }
 
-    // Adding custom states from webform js.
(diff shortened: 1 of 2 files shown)


anthony-malkoun (approved):
👍

vincent-gao (approved):

PR #529, opened 2024-10-21
[SD-115] Fixes conflicting field validations in textarea and textfield types

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-115

### Change
only apply the default `maxlength` to `textfield` type


yeniatencio (approved):
LGTM.

MdNadimHossain (approved):
@vincent-gao LGTM. I believe this has been suggested from Jo or Sarah. It will be best to request them to make a comment in the ticket about this change as they have raised it as a bug. So when we track back the change, we know the why it was changed. 
Thanks :)

PR #539, opened 2024-11-14
Added option to ignore service if not available.

MdNadimHossain:
### Problem/Motivation
The tide_core.system_info_service has a dependency on monitoring.sensor_runner service which creates the monitoring module dependency. The issue appears when the tide core service gets introduced first time including enabling the monitoring module. During the build step it tries to verify if the dependency service exists. But when the monitoring module getting enabled for the first time, during that build step it is still not enabled and service is not properly registered in the container. Because of this it ends up throwing a service dependency error.

Screenshot 2024-11-14 at 11 04 36 am

### Fix

1. Ignore the service if it is not defined.
2. Handle the null check the class.

Reference - https://eiriksm.dev/services-with-optional-dependencies-drupal-8
See the last option.


anthony-malkoun (approved):
Love the referenced article.

PR #556, opened 2024-11-27
[SD-535] Fixes character encoding strangeness

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-535

### Problem/Motivation
The character`Â` and other odd characters typically appear due to an encoding mismatch.

### Fix
Update `BasicTextEnhancer.php` to deal with this issue by explicitly specify the DOM will be UTF-8.

CleanShot 2024-11-27 at 14 18 08


MdNadimHossain (approved):
@vincent-gao did you get a chance to test the table properties and cell properties after this change to see if it still retains the style value that it suppose to do ?
may be do a quick check on that before merging this.

PR #562, opened 2024-12-11
[SD-551]AV-heading-not-visible

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SD-551
BE link: https://nginx-php.pr-1756.content-vic.sdp4.sdp.vic.gov.au/
### Problem/Motivation
Heading for AV is not fully visible:
image-20241203-031756

### Fix
Fixed styling by adding text-wrap property.
### Screenshots
Screenshot 2024-12-11 at 12 03 04 pm

### DIFF
--- a/modules/tide_paragraphs_enhanced_modal/css/paragraphs.enhanced_modal.css
+++ b/modules/tide_paragraphs_enhanced_modal/css/paragraphs.enhanced_modal.css
@@ -33,6 +33,7 @@
   box-shadow: none !important;
   border: none !important;
   padding: 0 0 0.5em;
+  text-wrap: auto;
 }
 
 .paragraphs-add-dialog-enhanced .paragraphs-add-dialog-list .paragraphs-add-dialog-row .description {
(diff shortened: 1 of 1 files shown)


modules/tide_paragraphs_enhanced_modal/css/paragraphs.enhanced_modal.css
```diff
@@ -33,6 +33,7 @@
   box-shadow: none !important;
   border: none !important;
   padding: 0 0 0.5em;
+  text-wrap: auto;
```
MdNadimHossain, line 36: @yeniatencio auto does not actually count as a valid value anymore and it is not a global value as well. In this case you should use a valid value for text-wrap it self. In this case use text-wrap: wrap; or text-wrap: balance; or text-wrap: pretty; or text-wrap: stable;

```
/* Keyword values */
text-wrap: wrap;
text-wrap: nowrap;
text-wrap: balance;
text-wrap: pretty;
text-wrap: stable;

/* Global values */
text-wrap: inherit;
text-wrap: initial;
text-wrap: revert;
text-wrap: revert-layer;
text-wrap: unset;
```

MdNadimHossain (changes requested):
@yeniatencio small change, please look into the comment

modules/tide_paragraphs_enhanced_modal/css/paragraphs.enhanced_modal.css
yeniatencio: Thank you @MdNadimHossain . I have updated my PR.

MdNadimHossain (approved):

PR #568, opened 2024-12-18
[SD-441]add secure file widget to entity browser

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SD-441
Testing link: https://nginx-php.pr-1760.content-vic.sdp4.sdp.vic.gov.au/
### Problem/Motivation
When adding media items to a content page that are not currently in the media library, the user has the ability to add media items to the media library at the same time. However it is not currently possible to do so for the new Secure File media items as the option is missing from the menu options in the select media embed pop up
### Fix
Add another Entity form with Secure file bundle from Add widget plugin select list.
Hook update to import config if the module is enabled.
### Screenshots

Screenshot 2024-12-18 at 11 32 09 pm

Screenshot 2024-12-18 at 11 32 24 pm

### DIFF
--- a/config/install/entity_browser.browser.tide_media_browser_iframe.yml
+++ b/config/install/entity_browser.browser.tide_media_browser_iframe.yml
@@ -69,3 +69,14 @@ widgets:
     weight: -9
     label: 'Add audio'
     id: entity_form
+  514a147d-5e8c-4006-8bdd-6959b6239548:
+    id: entity_form
+    uuid: 514a147d-5e8c-4006-8bdd-6959b6239548
+    label: 'Add secure file'
+    weight: 6
+    settings:
+      submit_text: 'Save secure file'
+      entity_type: media
+      bundle: secure_file
+      form_mode: default
+
--- a/config/install/entity_browser.browser.tide_media_browser_iframe.yml
+++ b/config/install/entity_browser.browser.tide_media_browser_iframe.yml
@@ -69,14 +69,3 @@ widgets:
     weight: -9
     label: 'Add audio'
     id: entity_form
-  514a147d-5e8c-4006-8bdd-6959b6239548:
-    id: entity_form
-    uuid: 514a147d-5e8c-4006-8bdd-6959b6239548
-    label: 'Add secure file'
-    weight: 6
-    settings:
-      submit_text: 'Save secure file'
-      entity_type: media
-      bundle: secure_file
-      form_mode: default
-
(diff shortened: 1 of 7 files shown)


modules/tide_media/tide_media.install
```diff
@@ -81,3 +81,35 @@ function tide_media_update_10005() {
   $config->set('track_enabled_target_entity_types', $result);
   $config->save();
 }
+
+/**
+ * Adds secure file widget to entity browser.
+ */
+function tide_media_update_10006() {
+  if (\Drupal::moduleHandler()->moduleExists('tide_media_secure_files')) {
```
anthony-malkoun, line 89: More a comment than anything, but would we normally do this like this or would we load the config from `config/install` and merge it in?

anthony-malkoun (approved):
This all looks pretty straightforward to me. Just a question on approach which shouldn't block the PR.

modules/tide_media/tide_media.install
```diff
@@ -81,3 +81,35 @@ function tide_media_update_10005() {
   $config->set('track_enabled_target_entity_types', $result);
   $config->save();
 }
+
+/**
+ * Adds secure file widget to entity browser.
+ */
+function tide_media_update_10006() {
```
MdNadimHossain, line 88: this should be in the` tide_media_secure_files` module instead of adding it here. Cause this should only exists if this module is enabled. So move this in the related module as a update hook and also put the same function in the install hook, so that it gets into the project when `tide_media_secure_files` is installed first time.

config/install/entity_browser.browser.tide_media_browser_iframe.yml
```diff
@@ -69,3 +69,14 @@ widgets:
     weight: -9
     label: 'Add audio'
     id: entity_form
+  514a147d-5e8c-4006-8bdd-6959b6239548:
```
MdNadimHossain, line 72: No need to add this here as this shouldn't be added if the module is enabled. If you put it in the tide_media_secure_files install hook, then every time it gets enabled, it will add this in this config.

MdNadimHossain (changes requested):
@yeniatencio function looks ok, just the placement needs to be changed. Please have a look into the comments

yeniatencio:
> @yeniatencio function looks ok, just the placement needs to be changed. Please have a look into the comments

Hey @MdNadimHossain , Thank you for your feedback. I have updated my PR for your review.

anthony-malkoun (approved):
👍

PR #570, opened 2024-12-24
Feature/sd 506 enable basic auth

sharmasahil:
### Jira
https://digital-vic.atlassian.net/browse/SD-507
### Problem/Motivation
Basic auth is disabled for BE non prods.
### Fix
1. Enabled shield in tide_core and added excluded paths as required in config
2. **To-do:** will need to update it in settings.php for non prod environments and apply to sites using ansible automation.
 1. and .


vincent-gao (approved):
hey @sharmasahil 
you may need to make adjustments using ansible,to conditionally turn off/on the shield based on the environments.
e.g. 
https://github.com/dpc-sdp/content-vic-gov-au/blob/develop/docroot/sites/default/envs/production/settings.php#L32-L33

PR #573, opened 2025-01-08
[SD-365] Node-Level Breadcrumb Management

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-365

### Changes
1. Implements a **Parent Pointer Tree** data structure stored in cache to track relationships for breadcrumbs, as it closely aligns with our scenario and requirements.  more details: https://en.wikipedia.org/wiki/Parent_pointer_tree
3. The jsonapi module will manage the rendering cache, while the `Parent Pointer Tree` will determine how and when to invalidate the cache for the node hierarchy.
CleanShot 2025-01-09 at 15 23 51

4. `::getChildrenTree` will also be used in the future to render the hierarchical tree on taxonomy pages.      
CleanShot 2025-01-09 at 15 21 07
5. provided a drush command to rebuild breadcrumb relationships.
 `drush rbr`
6. Adding a new content type will automatically include breadcrumb fields


vincent-gao:
lint shouldn't fail. I think we should adjust phpcs to include parameters, e.g. `--standard=Drupal,DrupalPractice --extensions=php,module,inc,install,test,profile,theme,css,info,yml`

PR #579, opened 2025-01-21
Fix for maxlength if counter type is set

sharmasahil:
### Jira
https://digital-vic.atlassian.net/browse/SD-616
### Problem/Motivation
Removed maxlength from submission data if counter type is set for word/character limit.


modules/tide_webform/tide_webform.module
```diff
@@ -388,6 +388,12 @@ function tide_webform_webform_submission_presave(WebformSubmission $webform_subm
     if (isset($element['#type']) && $element['#type'] == 'wizard_page') {
       return;
     }
+    // Remove maxlength if counter_type is set.
+    if (isset($element['#type']) && $element['#type'] == 'textarea') {
```
yeniatencio, line 392: Please update to  ===

yeniatencio (approved):
Just 1 small update.

PR #587, opened 2025-02-07
[SD-709]revert linkit in paragraph key journeys

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SD-709
Testing link: https://nginx-php.pr-502.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au/
### Problem/Motivation
Landing Page Header Links URL display change
### Fix
The cause of this issue was one config file that has been updated to linkit. Only the view form was needed it so I am reverting the form file and removing it from the hook update.
### Screenshots
Screenshot 2025-02-07 at 7 08 13 pm


anthony-malkoun (approved):
Good find Yeni!

PR #590, opened 2025-02-10
[SD-601]fixed conditional

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SD-601
### Problem/Motivation
Causing issues at the front end as there was a mistake in the conditional logic.
### Fix
Fixed conditional logic.


anthony-malkoun (approved):
Thanks for this Yeni!

PR #594, opened 2025-02-13
[SD-722] Adds path and status fields to media document form display.

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-722

### Change
Adds path and status fields to media document form display.


anthony-malkoun (approved):
Yep, this looks good. I suspect that installing media alias display does this so might have to make this hook fire after the hook that installs it (which I think is the previous hook in `.install`?)

PR #595, opened 2025-02-13
Update audit node command for D10

anthony-malkoun:
### Jira

https://digital-vic.atlassian.net/browse/SDPSUP-8990

### Problem/Motivation

D10 requires that all entity queries have an explicit access check. See https://www.drupal.org/node/3201242 for reference.

> For Drupal 10 this will be enforced by throwing an exception if ::accessCheck() is not called.

While there, the help text has been corrected and had some examples added because I forget how to use it every time :)

### Fix

Add the access check.

### DIFF
--- a/modules/tide_search/src/Commands/TideSearchCommands.php
+++ b/modules/tide_search/src/Commands/TideSearchCommands.php
@@ -19,11 +19,22 @@ use Drush\Commands\DrushCommands;
 class TideSearchCommands extends DrushCommands {
 
   /**
-   * Audit nodeids that needs to be published/indexed based on search index.
+   * Audits node ids that needs to be published/indexed based on search index.
+   * 
+   * The list of items not in the search index can be added using the following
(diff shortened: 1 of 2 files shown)


modules/tide_search/src/Commands/TideSearchCommands.php
```diff
@@ -91,14 +104,14 @@ public static function getNidsFromSearchIndex($indexId) {
       throw new ConsoleException(t('@index was not found'));
     }
     $total = $indexes[$indexId]->getTrackerInstance()->getTotalItemsCount();
-    $no_of_batches = ceil($total / 1000);
-    // If the result set is more than 1000 then run it in batch.
+    $no_of_batches = ceil($total / 500);
```
anthony-malkoun, line 107: Found 1000 was timing out on Elasticsearch so have wound it down a bit.

yeniatencio (approved):
LGTM.

PR #597, opened 2025-02-18
[SD-602]added geopoint class

yeniatencio:
### Jira
https://digital-vic.atlassian.net/browse/SD-602
BE link: https://nginx-php.pr-1779.content-vic.sdp4.sdp.vic.gov.au/
### Problem/Motivation
Adding the class GeoPoint so it can be reused in any project for latitude and longitude fields (maps) instead to repeat code.
### Fix
Create a class Geopoint with the function extractAndValidateLatLong() to validate values in latitude and longitude fields and return in the format needed for searchapi (latitude,longitude)
### Related PRs
https://github.com/dpc-sdp/content-vic-gov-au/pull/1779

### DIFF
--- a/src/GeoPoint.php
+++ b/src/GeoPoint.php
@@ -18,7 +18,7 @@ class GeoPoint {
    *   The node object to extract latitude and longitude from.
    *
    * @return string|bool
-   *   Returns the valid latitude, longitude as a string in the format "lat,long"
+   *   Returns the valid latitude, longitude as a string in format "lat,long"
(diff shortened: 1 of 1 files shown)


vincent-gao (changes requested):
hey @yeniatencio 
1. `tide_core` doesn't define the `field_latitude` and `field_longitude` fields, so this function doesn't belong here.
2. A dedicated class for verifying geo values seems excessive, especially when it contains only a single function.

PR #599, opened 2025-02-25
fix: unable to save Page title display style field

vincent-gao:
### Issue
The field can't be saved correctly, because if `#disabled = TRUE` whatever value is in the field will be ignored.

### Change
Use form validate instead.

### DIFF
--- a/modules/tide_landing_page/tide_landing_page.module
+++ b/modules/tide_landing_page/tide_landing_page.module
@@ -153,6 +153,7 @@ function tide_landing_page_form_node_form_alter(&$form, FormStateInterface $form
         }
       }
     }
+    $form['#validate'][] = 'tide_landing_page_header_style_node_form_validate';
   }
 
   // Add conditional field for show table of content.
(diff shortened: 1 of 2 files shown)


anthony-malkoun:
@vincent-gao can you explain how it works for new publications? Does this mean the behaviour is different from the original ticket? The user will need to submit the form to then get an error? Is there any reason we can't set the correct value and leave the field disabled?

vincent-gao:
Hi @anthony-malkoun , Thanks for the CR!
> can you explain how it works for new publications? 

could you please elaborate on what you mean? Users just need to create a publication as they normally would.

> Does this mean the behaviour is different from the original ticket?

Yes, the behavior is mostly the same. The only difference is that I removed the disabled field element.

> The user will need to submit the form to then get an error?

Users normally don't see an error. the validation is just preventing they accidentally change the value.
error example:
CleanShot 2025-02-26 at 09 24 52

>  Is there any reason we can't set the correct value and leave the field disabled?

Drupal docs says, #disabled: (bool) If TRUE, the element is shown but does not accept user input.
https://www.drupal.org/docs/drupal-apis/form-api/form-render-elements

anthony-malkoun (approved):
Thanks @vincent-gao, after discussing with you and Jess we can proceed with this to unblock the release. You're right that this should be reworked back into a standard Drupal field rather than the "virtual" field we have now but can be done with the adding header functionality to all content types work.

This actual code change is fine, but the approach is less than ideal but needs a bigger piece of rework.

PR #618, opened 2025-03-27
[SD-705] Add Link List paragraph type

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-750

### Problem/Motivation
1. Completed the changes required for SD-705.
2. Rewrote the`import` function. now it first tries to import configs from config/sync, then falls back to importing from the module.
3. Introduced the `repositionEntityFormDisplayFields` function. It helps developers programmatically move a field widget to a specific position while respecting `content regions` and `field groups`. **It works great, but still experimental, so use responsibly.**
4. Introduced `repositionComponentsRelativeToKey`, which helps developers rearrange weighted associative arrays. one use case is repositioning a paragraph component to a new spot. **It works great, but still experimental, so use responsibly.**
CleanShot 2025-03-31 at 11 51 36


tide_core.install
```diff
@@ -378,3 +378,41 @@ function tide_core_update_10012() {
...
+    'link_list' => 'paragraphs_type',
+    'paragraph.link_list.field_paragraph_link_list' => 'field_config',
+    'paragraph.link_list.field_paragraph_title' => 'field_config',
+    'paragraph.link_list.default' => 'entity_form_display',
+  ];
+  foreach ($configs as $name => $type) {
+    $storage = \Drupal::entityTypeManager()->getStorage($type);
+    if ($storage->load($name) !== NULL) {
+      continue;
+    }{
```
anthony-malkoun, line 398: These braces look a bit weird, probably not required.

src/TideEntityUpdateHelper.php
```diff
@@ -388,4 +390,186 @@ public function configMergeDeep($module_name, $config_directory, array $form_con
     }
   }
 
+  /**
+   * Inserts or repositions components relative to a reference key in an array.
+   *
+   * @example
+   *  $storage = Drupal::entityTypeManager()->getStorage('field_config')->load('node.landing_page.field_landing_page_component');
```
anthony-malkoun, line 397: Why not do the pre and post steps in the helper function?

tide_core.install
vincent-gao: The magic of the php compiler, even though I wrote it wrong, it still compiled just fine... I actually meant to write an if-else.

vincent-gao: 🤣

src/TideEntityUpdateHelper.php
vincent-gao: hey @anthony-malkoun This method mainly deals with associative arrays that include a `weight` property. It's not related to
`$storage = Drupal::entityTypeManager()->getStorage('field_config')->load('node.landing_page.field_landing_page_component');`
I included that line in the example just to show a possible use case.

vincent-gao: > Why not do the pre and post steps in the helper function?

Hey @anthony-malkoun Could you please clarify what you meant by this comment?

nicksantamaria (approved):
General approach to solve the story looks sane, config entities and naming convention looks fine.

Approving this with the caveat that I'm not familiar enough with the system to offer potential improvements to the implementation of the helper class and update hooks.

src/TideEntityUpdateHelper.php
anthony-malkoun: OK, got it @vincent-gao :). If a use case though is for modifying field_config, could there be another wrapper function we could use which could ALSO be a usage exampe? This wrapper would take a component name, e.g. `node.landing_page.field_landing_page_component` and something like the two field names and the position modifier?

vincent-gao: Hey @anthony-malkoun, thanks for explaining!
Just to elaborate a bit more, this helper method itself is not limited to modifying `field_config` settings specifically.
in your backend, there are many YML structures (not only field_config) following a similar pattern like:
```yaml
field_name_a:
  weight: 37
  other_prop: []
  other_prop_1: hello
field_name_b:
  weight: 5
  ...

```
Reordering or restructuring this type of associative array (with `weight` property) without it, doing reordering manually can be a bit messy. so the intention of this method is purely to provide a generic utility to handle this data structure more easily, regardless of the context.
The actual use case (whether it's field_config, block settings, or any other array structure) really depends on the developer's need.
I've updated the examples to reflect this better.

vincent-gao: Maybe the naming is a bit confusing. I think it probably needs some adjustment.

PR #620, opened 2025-03-31
removed unnecessary fields

yeniatencio:
### Jira
Testing link: https://nginx-php.pr-522.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au/
### Problem/Motivation
content-reference was down due to an outdated configuration file (field.field.node.publication_page.field_landing_page_component.yml), which was causing issues in the development environment. A quick fix to restore the reference was to push the updated configuration file to config/sync. However, the configuration file in config/install still required updates.
### Fix
Updated the configuration file and the corresponding update hook to ensure the file no longer has dependencies on news_listing and latest_events.

### Screenshots
Screenshot 2025-03-31 at 5 02 19 pm


vincent-gao (approved):
Thanks @yeniatencio .

PR #622, opened 2025-04-02
[SD-751] allow addition of top and bottom corner graphics for a site …

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-751

### Problem/Motivation
When top and bottom corner graphic files are added in the taxonomy of a site section, the images will flow through to all pages in that site section. If individual images are added at a node level, these will over-ride the site section images.

Add a custom field corner_graphic_field to the node that returns JSON in the following format:

e.g.
```json
    "corner_graphic_field": {
                "top_corner_graphic": "/sites/default/files/2025-04/8788145_0.jpeg",
                "bottom_corner_graphic": "/sites/default/files/2025-04/8788145.jpeg"
            },
```
 
 ```json
             "corner_graphic_field": {
                "top_corner_graphic": null,
                "bottom_corner_graphic": "/sites/default/files/2025-04/8788145.jpeg"
            },
```

```json
                "corner_graphic_field": {
                "top_corner_graphic": null,
                "bottom_corner_graphic": null
            },
```

### DIFF
--- a/src/Plugin/CornerGraphicField.php
+++ b/src/Plugin/CornerGraphicField.php
@@ -1,0 +1,158 @@
+<?php
+
+namespace Drupal\tide_core\Plugin;
+
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Field\FieldItemList;
+use Drupal\Core\TypedData\ComputedItemListTrait;
+use Drupal\node\NodeInterface;
+use Drupal\taxonomy\Entity\Term;
+
+/**
+ * Corner Graphic field implementation.
+ */
+class CornerGraphicField extends FieldItemList {
+
+  use ComputedItemListTrait;
+
+  /**
+   * Computes the field value.
+   */
+  protected function computeValue() {
+    $node = $this->getEntity();
+    if (!$node instanceof NodeInterface) {
+      return;
+    }
+    if (!\Drupal::moduleHandler()->moduleExists('tide_site')) {
+      return;
+    }
+    $site_id = \Drupal::request()->get('site');
+    /** @var \Drupal\tide_site\TideSiteHelper $siteHelper */
+    $siteHelper = \Drupal::service('tide_site.helper');
+    // Return can be null or term entity.
+    $primary_site_term_entity = $siteHelper->getEntityPrimarySite($node);
+    $file_url_generator = \Drupal::service('file_url_generator');
+
+    $sites = $siteHelper->getEntitySites($node);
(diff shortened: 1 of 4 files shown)


modules/tide_landing_page/tide_landing_page.module
```diff
@@ -605,3 +608,19 @@ function tide_landing_page_node_presave(NodeInterface $node) {
     }
   }
 }
+
+/**
+ * Implements hook_entity_base_field_info_alter().
+ */
+function tide_landing_page_entity_base_field_info_alter(&$fields, EntityTypeInterface $entity_type) {
+  if ($entity_type->id() == 'node') {
```
nicksantamaria, line 616: So this is in the tide_landing_page module, but the behavior applies to all node bundles?

modules/tide_landing_page/src/Plugin/CornerGraphicField.php
```diff
@@ -0,0 +1,141 @@
...
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The node entity.
+   * @param array $details
+   *   The details array to be modified.
+   * @param object $file_url_generator
+   *   The file URL generator service.
+   */
+  protected function addNodeCornerGraphics(NodeInterface $node, array &$details, $file_url_generator) {
+    if ($node->hasField('field_graphical_image') &&
```
nicksantamaria, line 80: Where is field_graphical_image added to the node bundles?

modules/tide_landing_page/tide_landing_page.module
vincent-gao: thank you @nicksantamaria for the CR. yes. this is not appropriate, the whole change should be moved to the `tide_site` submodule, since this PR also needs functionality from `tide_site`.

modules/tide_landing_page/src/Plugin/CornerGraphicField.php
vincent-gao: `tide_core` defines `field_graphical_image`, but whether to use it depends on the modules being enabled.

modules/tide_landing_page/tide_landing_page.module
anthony-malkoun: I agree with Nick here. I know we use "landing page" as a bit of a catch all, but we should be recognising this is common/shared functionality.

vincent-gao:
hey @nicksantamaria @anthony-malkoun 
I've moved the logic to tide_core since this function will be shared across other custom modules

nicksantamaria (approved):

PR #625, opened 2025-04-17
[SD-854] Restore the content category fields for all content types.

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-854

### Change
The field storage definition should be removed.


anthony-malkoun (approved):
Why was this even there in the first place?

PR #626, opened 2025-04-22
Only delete submissions marked as processed = 1

vincent-gao:
### Change
1. Remove error_log
2. Only delete submissions marked as processed = 1


anthony-malkoun (approved):
Makes sense!

PR #628, opened 2025-04-29
[SD-800] Fixes position for the field_topic field.

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-880

### Problem/Motivation
This PR mainly fixes a positioning issue for content-vic (and possibly other sites). I also saw it as a good opportunity to update some helper functions.

We are introducing a new `revert` function, which is essentially like running `drush cim`, but it updates the active configuration with the inactive configurations. The `import` function is also based on `drush cim`, but it’s used to import 
new configurations.

We are also updating the `repositionEntityFormDisplayFields` function by adding a new parameter that lets you specify whether you want to position the field _before_ or _after_ another field visually.

### DIFF
--- a/modules/tide_publication/tide_publication.install
+++ b/modules/tide_publication/tide_publication.install
@@ -290,7 +290,7 @@ function tide_publication_update_10009() {
   if (!isset($data['content']['field_topic']) || !isset($data['content']['field_content_category'])) {
     return;
   }
-  $data['content'] = $tideEntityUpdateHelper->repositionComponentsRelativeToKey($data['content'], ['field_topic' => $data['content']['field_topic']], 'field_content_category', 'before');
+  $data['content'] = $tideEntityUpdateHelper->repositionComponentsRelativeToKey($data['content'], ['field_topic' => $data['content']['field_topic']], 'field_content_category', TideEntityUpdateHelper::POSITION_BEFORE);
   $config->setData($data);
   $config->save();
 }
--- a/modules/tide_publication/tide_publication.install
+++ b/modules/tide_publication/tide_publication.install
@@ -278,3 +278,19 @@ function tide_publication_update_10008() {
       ->save();
   }
 }
+
+/**
+ * Update the position of field_topic for the publication type.
+ */
+function tide_publication_update_10009() {
+  /** @var \Drupal\tide_core\TideEntityUpdateHelper $tideEntityUpdateHelper */
+  $tideEntityUpdateHelper = \Drupal::service('tide_core.entity_update_helper');
+  $config = \Drupal::configFactory()->getEditable('core.entity_form_display.node.publication.default');
+  $data = $config->getRawData();
+  if (!isset($data['content']['field_topic']) || !isset($data['content']['field_content_category'])) {
+    return;
+  }
+  $data['content'] = $tideEntityUpdateHelper->repositionComponentsRelativeToKey($data['content'], ['field_topic' => $data['content']['field_topic']], 'field_content_category', 'before');
+  $config->setData($data);
+  $config->save();
+}
--- a/src/TideEntityUpdateHelper.php
+++ b/src/TideEntityUpdateHelper.php
@@ -28,6 +28,8 @@ class TideEntityUpdateHelper extends ConfigReverter {
 
   const INSTALL_DIR = '/config/install';
   const OPTIONAL_DIR = '/config/optional';
+  const POSITION_BEFORE = 'before';
+  const POSITION_AFTER = 'after';
 
   /**
    * The entity type manager.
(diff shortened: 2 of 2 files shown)


modules/tide_publication/tide_publication.install
```diff
@@ -278,3 +278,27 @@ function tide_publication_update_10008() {
...
+      'match_operator' => 'CONTAINS',
+      'size' => 60,
+      'placeholder' => '',
+      'match_limit' => 10,
+    ],
+    'third_party_settings' => [],
+    'type' => 'entity_reference_autocomplete',
+    'region' => 'content',
+  ];
+  $data['content'] = $tideEntityUpdateHelper->repositionComponentsRelativeToKey($data['content'], ['field_topic' => $field_topic], 'field_content_category', 'before');
```
vincent-gao, line 301: Here’s an example of how to use `repositionComponentsRelativeToKey`. Since it doesn’t involve grouped fields,  we can just use `repositionComponentsRelativeToKey` to position a field in an `entity_form_display` entity.

modules/tide_publication/tide_publication.install
```diff
@@ -278,3 +278,19 @@ function tide_publication_update_10008() {
...
+ */
+function tide_publication_update_10009() {
+  /** @var \Drupal\tide_core\TideEntityUpdateHelper $tideEntityUpdateHelper */
+  $tideEntityUpdateHelper = \Drupal::service('tide_core.entity_update_helper');
+  $config = \Drupal::configFactory()->getEditable('core.entity_form_display.node.publication.default');
+  $data = $config->getRawData();
+  if (!isset($data['content']['field_topic']) || !isset($data['content']['field_content_category'])) {
+    return;
+  }
+  $data['content'] = $tideEntityUpdateHelper->repositionComponentsRelativeToKey($data['content'], ['field_topic' => $data['content']['field_topic']], 'field_content_category', 'before');
```
vincent-gao, line 293: Here’s an example of how to use `repositionComponentsRelativeToKey`. Since it doesn’t involve grouped fields, we can just use `repositionComponentsRelativeToKey` to position a field in an `entity_form_display` entity.

src/TideEntityUpdateHelper.php
```diff
@@ -496,15 +496,15 @@ public function repositionComponentsRelativeToKey(array $components, array $new_
    *  ];
    *  $tide_helper->repositionEntityFormDisplayFields('node.landing_page.default', 'field_topic',['field_show_topic_term_and_tags'=>$field_show_topic_term_and_tags])
    */
-  public function repositionEntityFormDisplayFields(string $entity_form_display_id, string $after_field, array $fields_to_insert): void {
+  public function repositionEntityFormDisplayFields(string $entity_form_display_id, string $anchor_field, array $fields_to_insert, string $position = 'after'): void {
```
anthony-malkoun, line 499: It's a little thing, but if you made your options for `$position` class fields the IDE could auto complete?

src/TideEntityUpdateHelper.php
```diff
@@ -496,15 +496,15 @@ public function repositionComponentsRelativeToKey(array $components, array $new_
...
     $form_display = \Drupal::entityTypeManager()->getStorage('entity_form_display')->load($entity_form_display_id);
     if (!$form_display) {
       throw new \Exception("Form display not found: $entity_form_display_id");
     }
 
-    $anchor_component = $form_display->getComponent($after_field);
+    $anchor_component = $form_display->getComponent($anchor_field);
```
anthony-malkoun, line 505: Good naming here!

src/TideEntityUpdateHelper.php
```diff
@@ -608,4 +620,67 @@ public function import($type, $name, $prioritise_sync = TRUE) {
     return TRUE;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function revert($type, $name, $prioritise_sync = TRUE) {
```
anthony-malkoun, line 626: Could you explain a little bit what this is trying to do? I don't get why a revert function would `$prioritise_sync`?

anthony-malkoun: Does it try and fill in missing values from install config based on what is already there?

src/TideEntityUpdateHelper.php
vincent-gao: yep, Thanks! I will make this change !

src/TideEntityUpdateHelper.php
vincent-gao: By default, when `$prioritise_sync` is `true`, the `revert` function will first check  `/app/config/sync`. If the file isn't found there, it will look in the module’s `config/install`, and then in `config/optional`.

If `$prioritise_sync` is set to `false`, the function skips `/app/config/sync` and checks only `config/install`, then `config/optional`.

This makes it easier for developers to **update** specific config entities, for example, custom module defined YAML files in content-vic. The `revert` function handles UUIDs internally, so it won't cause any UUID changes.

anthony-malkoun (approved):
I think this is another good step to moving forward on our config management!

PR #629, opened 2025-05-01
[SDPSUP-9636] Unset field_published_date value on cloned nodes

nicksantamaria:
### Jira

https://digital-vic.atlassian.net/browse/SDPSUP-9636

### Problem/Motivation

Cloned nodes copy the value of `field_published_date` from the source entity. This field is not editable, which means the cloned node is stuck with whatever value was copied across.

This was a particular issue for Digital Public Notices who use that field for display on the front end.

### Fix

This fix unsets the value of `field_published_date` on the cloned entity.

### DIFF
--- a/tide_core.module
+++ b/tide_core.module
@@ -884,3 +884,12 @@ function tide_core_entity_base_field_info_alter(&$fields, EntityTypeInterface $e
       ->setCardinality(1);
   }
 }
+
+/**
+ * Implements hook_cloned_node_alter().
+ */
+function tide_core_cloned_node_alter(NodeInterface &$node, NodeInterface $original): void {
+  // Unsets the value of field_published_date on a cloned entity.
+  $node->set('field_published_date', NULL);
+  $node->save();
+}


tide_core.module
```diff
@@ -884,3 +884,11 @@ function tide_core_entity_base_field_info_alter(&$fields, EntityTypeInterface $e
       ->setCardinality(1);
   }
 }
+
+/**
+ * Implements hook_cloned_node_alter().
+ */
+function tide_core_cloned_node_alter(NodeInterface &$node, NodeInterface $original): void {
+  // Unsets the value of field_published_date on a cloned entity.
+  $node->set('field_published_date', NULL);
```
anthony-malkoun, line 893: Doesn't it need _a_ date? Should it be today's date?

nicksantamaria: For public notices this date is calculated when the node transitions to published state.

Its also on media document entities, which doesnt have the quick node clone feature, but also the default value is empty.

anthony-malkoun: I feel like removing the `isEmpty()` check is the lesser impact but I don't think anyone cares either way. Will this mean other cloned content types won't have a publish date?

nicksantamaria: >I feel like removing the isEmpty() check is the lesser impact

This would result in the published date being changed for every published revision. I dont think that is the intended behavior for DPN.

>Will this mean other cloned content types won't have a publish date?
Yep - I checked with Jo though and she thought that cloned content should not inherit the published date regardless of content type.

anthony-malkoun: Ah, that was my only concern. Good enough for me.

anthony-malkoun (approved):

vincent-gao (approved):

PR #632, opened 2025-05-06
[SD-903] Pin TFA to 1.9 for compatibility

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-903
### Problem/Motivation
- https://www.drupal.org/sa-contrib-2025-023
- https://git.drupalcode.org/project/tfa/-/compare/8.x-1.9...8.x-1.10?from_project_id=57554
CleanShot 2025-05-06 at 14 09 46

### Fix
To address compatibility issues introduced by drupal/tfa 1.10, this change pins the drupal/tfa dependency to version 1.9 in our composer.json. 
Upgrades to 1.10 will be rolled out as part of a future tide release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
         "drupal/smtp": "^1.2",
         "drupal/stage_file_proxy": "^2.0",
         "drupal/tablefield": "2.4",
-        "drupal/tfa": "^1.2",
+        "drupal/tfa": "1.9",
         "drupal/tfa_email_otp": "^1.0@beta",
         "drupal/token_conditions": "dev-compatible-with-d10",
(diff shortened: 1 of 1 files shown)


anthony-malkoun (approved):
Create a follow up ticket please to fix this.

vincent-gao:
> Create a follow up ticket please to fix this.

hey @anthony-malkoun ,thanks for the CR. 
 yes, Jess will create a ticket to align with the new rules and remove the current overrides. The goal is to make sure users still need to go through TFA after resetting their password, instead of being able to access the homepage right away. The current issue is that our code isn't compatible with version 1.10.

PR #635, opened 2025-05-09
Removed broken tide_event webform email handler

nicksantamaria:
### Jira

N/A

### Problem/Motivation

Currently most (if not all) tide projects are reporting issues with the tide_event webform not sending emails properly.

This is because the webform is configured out-of-the-box with a FROM address for some type of CRM solution that must have been [from the original implementation](https://github.com/dpc-sdp/tide_event/commit/5e5dbe47e129a907a30f056f050d59cbcd1aa067#diff-aa12a57db3b695989dde94fcf01846b45820a3cc134f37f59e00dc65a97b5ae8R275).

### Fix

There is no way to know specifically where each SDP tenant would want this to go, so the fix is just to remove the webform email handler all together.

### DIFF
--- a/modules/tide_event/config/install/webform.webform.tide_event_submission.yml
+++ b/modules/tide_event/config/install/webform.webform.tide_event_submission.yml
@@ -276,38 +276,4 @@ access:
       - authenticated
     users: {  }
     permissions: {  }
-handlers:
-  event_submission_email:
-    id: email
-    label: 'Event Submission Email'
-    handler_id: event_submission_email
-    status: true
-    conditions: {  }
-    weight: 0
-    settings:
-      states:
-        - completed
-      to_mail: contact.dpc@dsdbi.desk-mail.com
-      to_options: {  }
-      cc_mail: ''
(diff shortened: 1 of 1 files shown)


modules/tide_event/tide_event.install
```diff
@@ -106,3 +106,52 @@ function tide_event_update_10002() {
...
+  $updated = FALSE;
+
+  // Go through each handler.
+  foreach ($handlers as $handler_id => $handler) {
+    if ($handler_id == 'event_submission_email') {
+      $configuration = $handler->getConfiguration();
+
+      // Check if this is an email handler with the specific to_mail address.
+      if (isset($configuration['settings']['to_mail']) &&
+        $configuration['settings']['to_mail'] === 'contact.dpc@dsdbi.desk-mail.com') {
```
nicksantamaria, line 136: This is the key logic - if the webform is still configured to send an email to this address it should be removed.

vincent-gao (approved):
thanks @nicksantamaria . looks good to me

PR #641, opened 2025-05-21
Add environment indicator setting to display tag deployed

sharmasahil:
### Jira
https://digital-vic.atlassian.net/browse/SD-758
### Problem/Motivation
No indication of current tag deployed on non-prod/prod tide projects.
### Fix
Added config setting to provide relevant tag on tide projects.
### Related PRs
https://github.com/dpc-sdp/bay/pull/362

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -421,3 +421,23 @@ function tide_core_update_10015() {
   // Save the configuration.
   $config->save();
 }
+
+/**
+ * Add code deployed tag to environment.
+ */
+function tide_core_update_10016() {
+  $env = getenv('LAGOON_ENVIRONMENT_TYPE') ?: 'development';
+  $moduleHandler = \Drupal::service('module_installer');
+  if (!$moduleHandler->moduleExists('environment_indicator')) {
+    $moduleHandler->install(['environment_indicator']);
+  }
+
+  if ($env !== 'production') {
+    $config = \Drupal::configFactory()->getEditable('user.role.authenticated');
+    $permissions = $config->get('permissions') ?? [];
+    if (!in_array('access environment indicator', $permissions)) {
+      $permissions[] = 'access environment indicator';
+      $config->set('permissions', $permissions)->save();
+    }
+  }
+}


tide_core.install
```diff
@@ -421,3 +421,23 @@ function tide_core_update_10015() {
...
+ * Add code deployed tag to environment.
+ */
+function tide_core_update_10016() {
+  $env = getenv('LAGOON_ENVIRONMENT_TYPE') ?: 'development';
+  $moduleHandler = \Drupal::service('module_installer');
+  if (!$moduleHandler->moduleExists('environment_indicator')) {
+    $moduleHandler->install(['environment_indicator']);
+  }
+
+  if ($env !== 'production') {
```
vincent-gao, line 435: hey @sharmasahil 
I don't think we have a config split for user roles, so checking `$env !== 'production'` might not be necessary.

tide_core.install
```diff
@@ -421,3 +421,23 @@ function tide_core_update_10015() {
...
+    $config = \Drupal::configFactory()->getEditable('user.role.authenticated');
+    $permissions = $config->get('permissions') ?? [];
```
vincent-gao, line 437: To grant roles permissions, use [Role::grantPermission](https://api.drupal.org/api/drupal/core%21modules%21user%21src%21Entity%21Role.php/function/Role%3A%3AgrantPermission/10)

vincent-gao (changes requested):
hey @sharmasahil 
please also update the `user.role.authenticated.yml` file

tide_core.install
sharmasahil: This is just config update to add one more permission to roles in non prod.

sharmasahil:
> hey @sharmasahil please also update the `user.role.authenticated.yml` file

Update with??

tide_core.install
vincent-gao: hey @sharmasahil , You want the update hook to run only in non-prod environments, right? But in reality, this setting will always be exported. e.g. when we run an automation upgrade, it’ll still get deployed to production. We don’t maintain 2 separate databases to keep prod and non-prod configs, and we don't mange roles in different environments . Plus, update hooks only run once, so your setting would still get overwritten by the PROD config.

vincent-gao:
> > hey @sharmasahil please also update the `user.role.authenticated.yml` file
> 
> Update with??

Right, it's defined by core. All good. Could you also include the permission change in the `tide_core_install` hook to make sure it's applied during the tide_core installation?

tide_core.install
anthony-malkoun: Why do we want it to run in non prod only @sharmasahil / @vincent-gao? This seems useful for production as well?

anthony-malkoun: Ah, I just read the AC of the JIRA ticket. Carry on.

PR #645, opened 2025-05-26
[SD-898] Added approver plus roles and restrictions.

MdNadimHossain:

### DIFF
--- a/modules/tide_site_restriction/tide_site_restriction.module
+++ b/modules/tide_site_restriction/tide_site_restriction.module
@@ -139,23 +139,23 @@ function tide_site_restriction_views_pre_view(ViewExecutable $view, $display_id,
       'media_page_list' => 'field_media_site_target_id',
     ],
   ];
-
+  
   $account = \Drupal::currentUser();
   $view_id = $view->id();
 
   // Apply site restriction filters for supported views.
   if (isset($display_filters[$view_id][$display_id])) {
     $filter = $display_filters[$view_id][$display_id];
-
+  
     /** @var \Drupal\tide_site_restriction\Helper $site_restriction_helper */
     $site_restriction_helper = \Drupal::service('tide_site_restriction.helper');
     $user_can_bypass_restriction = $site_restriction_helper->canBypassRestriction($account);
-
+  
     $filters = $view->display_handler->getOption('filters');
     if (!empty($filters[$filter])) {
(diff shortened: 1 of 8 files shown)


modules/tide_site_restriction/tide_site_restriction.module
```diff
@@ -394,3 +398,39 @@ function tide_site_restriction_views_data_alter(array &$data) {
 function tide_site_restriction_views_plugins_field_alter(array &$plugins) {
   $plugins['node_bulk_form']['class'] = 'Drupal\\tide_site_restriction\\Plugin\\views\\field\\TideNodeBulkForm';
 }
+
+/**
+ * Implements hook_field_widget_single_element_form_alter().
+ */
+function tide_site_restriction_field_widget_single_element_form_alter(&$element, FormStateInterface $form_state, $context) {
```
MdNadimHossain, line 405: this function is to show all the sites and if there is any child site under it during editing other users.

modules/tide_ui_restriction/tide_ui_restriction.module
```diff
@@ -75,3 +77,31 @@ function tide_ui_restriction_admin_config_access_check_exclude($path) : bool {
...
+function tide_ui_restriction_form_user_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  $current_user = \Drupal::currentUser();
+  if (!$current_user->hasRole('approver_plus')) {
+    return;
+  }
+
+  // Hide higher roles.
+  if (isset($form['account']['roles']) && is_array($form['account']['roles']['#options'])) {
+    unset($form['account']['roles']['#options']['administrator']);
+    unset($form['account']['roles']['#options']['site_admin']);
```
MdNadimHossain, line 93: Approver plus can not assign higher roles.

modules/tide_ui_restriction/tide_ui_restriction.permissions.yml
```diff
@@ -4,3 +4,12 @@ access admin_config pages:
 'access event log track':
   title: 'Access event log track'
   description: 'Allow user to access event log track.'
+access user creation page:
```
anthony-malkoun, line 7: Was this the bit that was needed to allow user creation? i.e. the permission was fine but we were restricting the route?

sharmasahil (approved):

PR #646, opened 2025-05-27
[SD-964] YAML enhancer supports tokens

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-964

### Problem/Motivation
The token `[current-date:html_date]` should populate the date field with the current date when the webform is loaded.

### Fix
Inlcude the token service

### DIFF
--- a/modules/tide_api/src/Plugin/jsonapi/FieldEnhancer/YamlEnhancer.php
+++ b/modules/tide_api/src/Plugin/jsonapi/FieldEnhancer/YamlEnhancer.php
@@ -24,12 +24,21 @@ class YamlEnhancer extends ResourceFieldEnhancerBase {
   protected function doUndoTransform($data, Context $context) {
     $data = Yaml::decode($data);
 
+    // Include the token service to process any tokens in the YAML data.
+    $token_service = \Drupal::service('token');
     if (!empty($data['processed_text']['#text'])) {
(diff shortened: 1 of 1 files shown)


modules/tide_api/src/Plugin/jsonapi/FieldEnhancer/YamlEnhancer.php
```diff
@@ -24,13 +24,22 @@ class YamlEnhancer extends ResourceFieldEnhancerBase {
...
+    // Include the token service to process any tokens in the YAML data.
+    $token_service = \Drupal::service('token');
     if (!empty($data['processed_text']['#text'])) {
       $data['processed_text']['#text'] = $this->processText($data['processed_text']['#text']);
     }
 
     if (!empty($data['markup']['#markup'])) {
       $data['markup']['#markup'] = $this->processText($data['markup']['#markup']);
     }
+    // Process any other fields that may contain token replacements.
```
vincent-gao, line 36: This way only decodes the tokens that the `drupal/token`and `drupal/core` itself knows about. It has no dependency on `webform` or the  `submission` entity, so it cannot process any tokens defined by the `drupal/webform` module. As a result, generic tokens, such as date, language, site name, etc.—will work fine, but any `webform` defined tokens will not be recognised.

MdNadimHossain (approved):

PR #650, opened 2025-05-29
remove patch jsoapi menu items

yeniatencio:
### Problem/Motivation
Yesterday a new release for JSON:API Menu Items has been launched and this [patch](https://www.drupal.org/project/jsonapi_menu_items/issues/3350524#comment-15577964) is not needed anymore. 
### Screenshots
Screenshot 2025-05-30 at 9 07 40 am
Screenshot 2025-05-30 at 9 06 52 am

PR #652, opened 2025-06-02
Re-rolled patch to resolve the conflict with v 1.2.7.

MdNadimHossain:
### Problem/Motivation
The patch was removed in the is https://github.com/dpc-sdp/tide_core/pull/650. But it should have been re-rolled and updated. 
### Fix
Added back after re-rolling the patch with the new version 1.2.7.


yeniatencio (approved):
Thanks Nadim. I have should looked that new release contains the fix.

PR #654, opened 2025-06-03
[SD-995] Added field check before trying to set the field as null.

MdNadimHossain:
### Jira
https://digital-vic.atlassian.net/browse/SD-995
### Problem/Motivation
During the cloning process, certain fields may not be present on all content types. Attempting to set a value—such as NULL—on a field that doesn't exist causes a fatal error.
### Fix
1. Added a a simple check if node has the field and then set it as null.


anthony-malkoun (approved):
👍

PR #659, opened 2025-06-11
[SD-849] update widget site field 

yeniatencio:

### DIFF
--- a/modules/tide_alert/config/install/core.entity_form_display.node.alert.default.yml
+++ b/modules/tide_alert/config/install/core.entity_form_display.node.alert.default.yml
@@ -4,6 +4,7 @@ dependencies:
   config:
     - field.field.node.alert.field_alert_type
     - field.field.node.alert.field_call_to_action
+    - field.field.node.alert.field_node_site
     - node.type.alert
   module:
     - content_moderation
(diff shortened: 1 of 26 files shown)


vincent-gao (changes requested):
Hey @yeniatencio, field_nod_site is defined programmatically by the tide_site module via [this function](https://github.com/dpc-sdp/tide_core/blob/develop/modules/tide_site/tide_site.module#L33). So it shouldn't be added through YAML, it's handled directly in the `tide_site` module.

vincent-gao (changes requested):
hey @yeniatencio , unfortunately, this still isn’t the right way to do it.
basicly just as what I mentented in previous comment, 
> Hey @yeniatencio, field_nod_site is defined programmatically by the tide_site module via [this function](https://github.com/dpc-sdp/tide_core/blob/develop/modules/tide_site/tide_site.module#L33). So it shouldn't be added through YAML, it's handled directly in the tide_site module.

Please cover two scenarios:

- Adding a new bundle

  Tweak the function in [tide_site.module](https://github.com/dpc-sdp/tide_core/blob/develop/modules/tide_site/tide_site.module#L33) so any new bundle gets this setting automatically.

- Updating existing bundles

   Apply the widget to all current bundles

PR #660, opened 2025-06-13
[Release 1.54.0 issues] [SD-1027] Fix relation issue from search_api_exclude_entity

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-1027

### Issues
This PR addresses the issues identified during the release.

### Changes
1.  Move search_api to the tide_core level to fix dependency issues when submodules use it in their logic.
2. Add logics to `hook_update_dependencies` to ensure the update hooks run in a currect order.
3. Update tide_site_entity_presave to handle the user entity form case.

### DIFF
--- a/modules/tide_landing_page/tide_landing_page.install
+++ b/modules/tide_landing_page/tide_landing_page.install
@@ -37,7 +37,7 @@ function tide_landing_page_update_dependencies() {
   $dependencies['tide_landing_page'][10101] = ['tide_core' => 10005];
   $dependencies['tide_landing_page'][10106] = ['tide_core' => 10009];
   $dependencies['tide_landing_page'][10111] = ['tide_core' => 10014];
-
+  $dependencies['tide_landing_page'][10113] = ['tide_core' => 10018];
   return $dependencies;
 }
 
@@ -361,13 +361,13 @@ function tide_landing_page_update_10112() {
  * Add search exclude field.
  */
 function tide_landing_page_update_10113() {
-  $update_service = \Drupal::service('tide_core.entity_update_helper');
-  // Import search exclude fields.
-  $field_configs = [
-    'field.field.node.landing_page.field_search_index_exclude' => 'field_config',
-  ];
-  foreach ($field_configs as $name => $type) {
-    $update_service->import($type, $name);
+  \Drupal::moduleHandler()->loadInclude('tide_core', 'inc', 'includes/helpers');
+  $config_location = [\Drupal::service('extension.list.module')->getPath('tide_landing_page') . '/config/install'];
+  $config_read = _tide_read_config('field.field.node.landing_page.field_search_index_exclude', $config_location, TRUE);
+  $storage = \Drupal::entityTypeManager()->getStorage('field_config');
+  if ($storage->load('node.landing_page.field_search_index_exclude') == NULL) {
+    $config_entity = $storage->createFromStorageRecord($config_read);
+    $config_entity->save();
   }
 
   $form_display = \Drupal::entityTypeManager()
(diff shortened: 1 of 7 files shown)


vincent-gao:
> Hey @vincent-gao re:
> 
> > Move search_api to the tide_core level to fix dependency issues when submodules use it in their logic.
> 
> Were other modules besides `tide_search` using Search API?

Yes, this PR actually re-fixes the issue reported in https://digital-vic.atlassian.net/browse/SD-665. Both `tide_landing_page` and `tide_publication` try to set the `field_search_index_exclude` field, which is defined by the `drupal/search_api_exclude_entity` field type. However, neither of these modules has a direct dependency on `drupal/search_api_exclude_entity`. By moving the `search_api` and `search_api_exclude_entity` into tide_core, we make the relationship meaningful without breaking the install process. tide_search still manages the actual search API/elasticsearchc configuration.

anthony-malkoun:
> > Hey @vincent-gao re:
> > > Move search_api to the tide_core level to fix dependency issues when submodules use it in their logic.
> > 
> > 
> > Were other modules besides `tide_search` using Search API?
> 
> Thanks for the CR @anthony-malkoun . Yes, this PR actually re-fixes the issue reported in https://digital-vic.atlassian.net/browse/SD-665. Both `tide_landing_page` and `tide_publication` try to set the `field_search_index_exclude` field, which is defined by the `drupal/search_api_exclude_entity` field type. However, neither of these modules has a direct dependency on `drupal/search_api_exclude_entity`. By moving the `search_api` and `search_api_exclude_entity` into tide_core, we make the relationship meaningful without breaking the install process. tide_search still manages the actual search API/elasticsearchc configuration. tide_core is just simply enabling the `search_api` and `search_api_exclude_entity` module.

Hey @vincent-gao I'm not against the approach, but we have similar situation with `tide_site` don't we? There are hooks in there that add the site fields to new content types as we add them. Shouldn't we do it like that here too?

vincent-gao:
> > > Hey @vincent-gao re:
> > > > Move search_api to the tide_core level to fix dependency issues when submodules use it in their logic.
> > > 
> > > 
> > > Were other modules besides `tide_search` using Search API?
> > 
> > 
> > Thanks for the CR @anthony-malkoun . Yes, this PR actually re-fixes the issue reported in https://digital-vic.atlassian.net/browse/SD-665. Both `tide_landing_page` and `tide_publication` try to set the `field_search_index_exclude` field, which is defined by the `drupal/search_api_exclude_entity` field type. However, neither of these modules has a direct dependency on `drupal/search_api_exclude_entity`. By moving the `search_api` and `search_api_exclude_entity` into tide_core, we make the relationship meaningful without breaking the install process. tide_search still manages the actual search API/elasticsearchc configuration. tide_core is just simply enabling the `search_api` and `search_api_exclude_entity` module.
> 
> Hey @vincent-gao I'm not against the approach, but we have similar situation with `tide_site` don't we? There are hooks in there that add the site fields to new content types as we add them. Shouldn't we do it like that here too?

@anthony-malkoun To be honest, that was my first thought too! If we make the field available to all content types, it simplifies things. But since the SD and AC might change, should I check with Jess to confirm?

vincent-gao:
hey @anthony-malkoun I had a chat with Jess. Since we are on a tight timeline, we won't update the SD and AC for now. Jess will raise a new ticket to restructure the dependency and add the feature to `tide_search`, similar to how it's done with `tide_site`. so the exclude_field gets added by default to all bundles, and new bundles.

vincent-gao:
the ticket https://digital-vic.atlassian.net/browse/SD-1028

PR #674, opened 2025-07-23
Re-rolled webform patch with the new webform release.

MdNadimHossain:
### Problem/Motivation
The patch is outdated with the new release.
### Fix
Re-rolled the patch with the new webform release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -392,8 +392,7 @@
                 "Automated Drupal 10 compatibility fixes - https://www.drupal.org/project/jira_rest/issues/3288088#comment-14567776": "https://www.drupal.org/files/issues/2022-06-16/jira_rest.4.x-dev.rector.patch"
             },
             "drupal/webform": {
-                "Exporting webform submission as batch does not allowed for extended field due to static method - https://www.drupal.org/project/webform/issues/3348336#comment-16202432": "https://www.drupal.org/files/issues/2025-07-23/exporting-webform-submission-static-batch-process-3348336-9.patch",
-                "Fix incompatibility with devel ^5.2 - https://www.drupal.org/project/webform/issues/3524877#comment-16111172": "https://www.drupal.org/files/issues/2025-05-16/webform-6.3.0-beta2-3524877.patch"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -392,8 +392,7 @@
                 "Automated Drupal 10 compatibility fixes - https://www.drupal.org/project/jira_rest/issues/3288088#comment-14567776": "https://www.drupal.org/files/issues/2022-06-16/jira_rest.4.x-dev.rector.patch"
             },
             "drupal/webform": {
-                "Exporting webform submission as batch does not allowed for extended field due to static method - https://www.drupal.org/project/webform/issues/3348336#comment-15686345": "https://www.drupal.org/files/issues/2024-07-24/exporting-webform-submission-static-batch-process-3348336-8.patch",
-                "Fix incompatibility with devel ^5.2 - https://www.drupal.org/project/webform/issues/3524877#comment-16111172": "https://www.drupal.org/files/issues/2025-05-16/webform-6.3.0-beta2-3524877.patch"
```
MdNadimHossain, line 396: the return type is fixed and it is expecting a array return now. So need to use this patch where it declares the function as void. See the updated function here - https://git.drupalcode.org/project/webform/-/blob/6.3.x/src/Plugin/DevelGenerate/WebformSubmissionDevelGenerate.php?ref_type=heads#L258

anthony-malkoun: Do you mean "no need to use this patch"?

composer.json
```diff
@@ -392,8 +392,7 @@
...
+                "Exporting webform submission as batch does not allowed for extended field due to static method - https://www.drupal.org/project/webform/issues/3348336#comment-16202432": "https://www.drupal.org/files/issues/2025-07-23/exporting-webform-submission-static-batch-process-3348336-9.patch"
```
anthony-malkoun, line 395: Regarding this patch, as I see it, it will never get merged in as it is specific to our field. Can this be extended via a hook instead?

composer.json
MdNadimHossain: yes, looks like we don't need to use it. Cause the patch says it was missing the void type. But now it actually expects a array return after the new change.

anthony-malkoun (approved):
Always good to see a patch removed 👍. @MdNadimHossain as discussed let's create a follow up ticket to try and remove the other patch. It will need to be more generic and not reference our actual field.

PR #679, opened 2025-08-01
Feature/sd 1055 make default index shard count configurable

MdNadimHossain:
### Jira
https://digital-vic.atlassian.net/browse/SD-1055
### Fix
1. Override the class to add the number of shards in the UI and use that value as default 1. 
2. In the custom create function if the config value is not available set it as 1 by default.
3. This value can be editable easily using the UI and export it in the configs.


anthony-malkoun (approved):
This is a great PR. I like how it takes different edge cases into account and allows it to be set via the UI.

nicksantamaria (approved):

vincent-gao (approved):

PR #681, opened 2025-08-05
[SD-914] refactoring header style field.

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-914
https://digital-vic.atlassian.net/browse/SD-22

### Problem/Motivation
The existing header_style field logic has grown tangled over time, with overlapping conditions. This PR cleans up that mess by removing dead code, consolidating the decision flow, and making the mapping between Header Variant and Text Style Option explicit and automatic.

Header Variant | Text Style Option
-- | --
Default | Default OR Blocked
Corner Graphics | Default OR Blocked
Full BG image | Blocked [automatic]
Half BG Image / CTA | Default [automatic]

content-vic 
  - https://github.com/dpc-sdp/content-vic-gov-au/pull/1898

### DIFF
--- a/config/install/editor.editor.rich_text.yml
+++ b/config/install/editor.editor.rich_text.yml
@@ -77,15 +77,15 @@ settings:
         - '<dt>'
         - '<dd>'
         - '<footer>'
-        - '<span class>'
+        - '<span>'
         - '<img src alt data-entity-type data-entity-uuid data-align data-caption>'
-        - '<div class>'
+        - '<div class="quotation button button--secondary">'
         - '<iframe frameborder height scrolling src width title allowfullscreen>'
-        - '<drupal-entity data-caption data-align data-entity-type data-entity-uuid data-embed-button data-entity-embed-display data-entity-embed-display-settings data-show-last-updated title alt>'
-        - '<a hreflang target data-entity-substitution data-entity-type data-entity-uuid>'
-        - '<blockquote cite class>'
-        - '<ul type class>'
-        - '<ol type class>'
+        - '<drupal-entity data-entity-type data-embed-button data-show-last-updated>'
+        - '<a hreflang target>'
+        - '<blockquote cite>'
+        - '<ul>'
+        - '<ol type>'
         - '<h2 id>'
         - '<h3 id>'
         - '<h4 id>'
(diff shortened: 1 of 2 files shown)


vincent-gao:
> @vincent-gao if I read this right, it is only available for landing pages now? Isn't it on publications?

Hi @anthony-malkoun , it still works for publication. https://github.com/dpc-sdp/tide_core/blob/69abdb32f83550bec6ef3445c9d7f83c7ffee32d/modules/tide_publication/tide_publication.module#L135
1. The biggest change is how the pseudo field `header_style` is stored. Its value is now saved to the DB using the State API, whereas previously it was derived through a complex,  unreliable and wrong calculation.
2. Another change is the AJAX callback, which now fully meets the requirements.

tide_core.module
```diff
@@ -919,6 +919,12 @@ function tide_core_cloned_node_alter(NodeInterface &$node, NodeInterface $origin
   if ($node->hasField('field_published_date')) {
     $node->set('field_published_date', NULL);
   }
+  // Look up the original node's stored header style.
+  $header_style = \Drupal::state()->get($original->uuid() . '-header_style');
+  if ($header_style) {
+    // Store the same header style under the new node's UUID.
+    \Drupal::state()->set($node->uuid() . '-header_style', $header_style);
+  }
```
MdNadimHossain, line 927: this will make sure to copy over the header style option from the previous node to the cloned node

modules/tide_landing_page/tide_landing_page.module
```diff
@@ -530,8 +530,9 @@ function _tide_landing_page_form_node_form_process(array $form, FormStateInterfa
       'callback' => 'tide_landing_page_set_value_hero_image_theme',
     ],
   ];
-    $header_style = \Drupal::state()->get($node->uuid() . '-' . 'header_style');
-    $form['_header_style']['_header_style_options']['#default_value'] = $node->isNew() ? 'default' : $header_style;
+    $header_style = \Drupal::state()->get($node->uuid() . '-header_style');
+    $form['_header_style']['_header_style_options']['#default_value'] = $header_style ?: 'default';
```
MdNadimHossain: the new node can be cloned node as well, so this check will make sure if there is any value set in drupal state that comes from the original to the new cloned node. If it does then it will move the original node value to the clone node as well. If it is empty then it will set it as default.

MdNadimHossain (approved):

PR #689, opened 2025-09-16
[SD-1194] Added webform elements enhancer.

MdNadimHossain:
### Jira
https://digital-vic.atlassian.net/browse/SD-1194
### Problem/Motivation
The Advanced HTML/Text needs to be processed for the FE to render images.
### Fix
Added a element enhancer just like basic text enhancer. The structure for webform element data is different as it returns all the elements. This is why the enhancer needs to be added to the elements unlike where the basic text enhancer gets added to a field. For this reason it is better to create a separate enhancer and manipulate the elements. If we need any other manipulation, we can just add function to customise the same enhancer for different elements.

### DIFF
--- a/modules/tide_webform/config/optional/jsonapi_extras.jsonapi_resource_config.webform--webform.yml
+++ b/modules/tide_webform/config/optional/jsonapi_extras.jsonapi_resource_config.webform--webform.yml
@@ -125,7 +125,7 @@ resourceFields:
     fieldName: elements
     publicName: elements
     enhancer:
-      id: default_maxlength_enhancer
+      id: webform_elements_enhancer
     disabled: false
     advancedOptionsIcon: ''
     enhancer_label: ''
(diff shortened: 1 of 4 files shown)


modules/tide_webform/config/optional/jsonapi_extras.jsonapi_resource_config.webform--webform.yml
```diff
@@ -125,7 +125,7 @@ resourceFields:
     fieldName: elements
     publicName: elements
     enhancer:
-      id: default_maxlength_enhancer
```
MdNadimHossain, line 128: this enhancer is not in use anymore and I have checked all the project. But still kept the code so if we need this in future.

modules/tide_webform/tests/behat/features/jsonapi.content_rating.feature
```diff
@@ -15,7 +15,6 @@ Feature: JSON API Webform
     And the JSON node "data[0].attributes.elements.url" should exist
     And the JSON node "data[0].attributes.elements.was_this_page_helpful" should exist
     And the JSON node "data[0].attributes.elements.comments" should exist
-    And the JSON node "data[0].attributes.elements.comments.#maxlength" should be equal to "255"
```
MdNadimHossain, line 18: Removed the test for the unused enhancer.

vincent-gao (approved):

PR #690, opened 2025-09-16
[SD-1155] apply token to wizard pages

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-1155

### Problem/Motivation
Apply the token service to the wizard pages.


vincent-gao:
Update the logic to process token replacements recursively, so it can handle potential new types of nested arrays in the future without needing further changes.

PR #695, opened 2025-09-25
[SD-902] Revoke delete from all roles

waitingallday:
### Jira

https://digital-vic.atlassian.net/browse/SD-902

### Problem/Motivation

Revoke delete permission from all user roles.

### DIFF
--- a/tide_core.install
+++ b/tide_core.install
@@ -715,29 +715,19 @@ function tide_core_update_10022() {
 }
 
 /**
- * Revoke delete from all content types on all roles except administrator.
+ * Revoke `delete` from all content types on the 3 roles.
  */
 function tide_core_update_10023() {
-  $helper = \Drupal::service('tide_content_collection.search_api.index_helper');
-  $node_types = $helper->getNodeTypes();
-  $content_types = array_keys($node_types);
-
-  $roles = Role::loadMultiple();
-  foreach ($roles as $role_id => $role) {
-    if ($role_id === "administrator") {
-      continue;
-    }
-    $changed = FALSE;
-    $permissions = $role->getPermissions();
-    foreach ($permissions as $permission) {
-      foreach ($content_types as $type) {
-        if ($permission === "delete own {$type} content" || $permission === "delete any {$type} content") {
-          $role->revokePermission($permission);
-          $changed = TRUE;
-        }
-      }
-    }
-    if ($changed) {
+  $content_types = \Drupal::entityTypeManager()->getStorage('node_type')->loadMultiple();
+  $roles = [
+    'editor',
+    'approver',
+    'approver_plus',
+  ];
+  foreach ($roles as $role_id) {
+    $role = Role::load($role_id);
+    foreach ($content_types as $type => $content_type_entity) {
+      $role->revokePermission("delete own {$type} content")->revokePermission("delete any {$type} content");
       $role->save();
     }
   }


config/install/user.role.approver.yml
```diff
@@ -42,7 +42,6 @@ permissions:
   - 'create paragraph library item'
   - 'edit paragraph library item'
   - 'delete all revisions'
-  - 'delete any media'
```
MdNadimHossain, line 45: it only requires to remove the node permissions not the media, so no need to remove this.

config/install/user.role.approver_plus.yml
```diff
@@ -59,16 +59,7 @@ permissions:
...
-  - 'delete any audio media'
-  - 'delete any document media'
-  - 'delete any embedded_video media'
-  - 'delete any file media'
-  - 'delete any image media'
-  - 'delete any media'
   - 'delete media'
-  - 'delete own audio media'
-  - 'delete own document media'
-  - 'delete own image media'
```
MdNadimHossain, line 71: same no need to remove all these, as the ticket says only to remove node permissions

config/install/user.role.site_admin.yml
```diff
@@ -77,11 +77,6 @@ permissions:
   - 'create terms in topic'
   - 'create url aliases'
   - 'delete all revisions'
-  - 'delete any audio media'
-  - 'delete any document media'
-  - 'delete any embedded_video media'
-  - 'delete any file media'
-  - 'delete any image media'
```
MdNadimHossain, line 84: same here

modules/tide_media/modules/tide_media_secure_files/config/install/user.role.secure_file_user.yml
```diff
@@ -13,9 +13,6 @@ is_admin: null
 permissions:
   - 'create field_secure_file'
   - 'create secure_file media'
-  - 'delete any secure_file media'
-  - 'delete any secure_file media revisions'
-  - 'delete own secure_file media'
```
MdNadimHossain, line 18: same here

tide_core.install
```diff
@@ -713,3 +713,26 @@ function tide_core_update_10022() {
...
+      if (preg_match("/^delete (own|any) /", $permission)) {
+        $role->revokePermission($permission);
+        $changed = TRUE;
+      }
+    }
+    if ($changed) {
+      $role->save();
+    }
+  }
+}
```
MdNadimHossain, line 738: this update hook will revoke any other permission that has delete own or delete any which includes media, webform etc. Which is not the requirement for this ticket. 
The approach should be, load the content types of a CMS e.g content-vic or content-legislation. This will load all the custom content types as well. Then remove only the delete permissions for all the content types available for that CMS. The other permissions should be unchanged. 
There is [this helper function here](https://github.com/dpc-sdp/tide_core/blob/develop/modules/tide_api/modules/tide_content_collection/src/SearchApiIndexHelper.php#L139) that you can use to load the content types or you can re-create the same logic in the update hook to load the content types. 
```
$helper = \Drupal::service('tide_content_collection.search_api.index_helper');
  $node_types = $helper->getNodeTypes();
  $content_types = array_keys($node_types);
```
Then target the permissions for that content type like this - 
 ```
$changed = FALSE;
    $permissions = $role->getPermissions();
    foreach ($permissions as $permission) {
      foreach ($content_types as $type) {
        if ($permission === "delete own $type" || $permission === "delete any $type") {
          $role->revokePermission($permission);
          $changed = TRUE;
        }
      }
    }
    if ($changed) {
      $role->save();
    }
```

MdNadimHossain (changes requested):
@waitingallday please have a look into the requested changes and check if it make sense. Let me know if you have any questions.

MdNadimHossain (approved):

MdNadimHossain:
LGTM. Please make sure the update hook works as expected by testing it.

PR #697, opened 2025-09-30
[SD-1156] Tags Relocation

vincent-gao:

### DIFF
--- a/modules/tide_site/tide_site.module
+++ b/modules/tide_site/tide_site.module
@@ -510,13 +510,12 @@ function tide_site_form_node_form_alter(&$form, FormStateInterface $form_state) 
     '#open' => $node->isNew(),
     '#optional' => TRUE,
     '#access' => TRUE,
-    '#group' => 'advanced',
+    '#group' => 'group_page_metadata',
     '#attributes' => [
       'class' => ['tide-site-metadata'],
     ],
     '#weight' => 100,
   ];
-
   $site_fields = [
     0 => TideSiteFields::normaliseFieldName(TideSiteFields::FIELD_PRIMARY_SITE, 'node'),
     10 => TideSiteFields::normaliseFieldName(TideSiteFields::FIELD_SITE, 'node'),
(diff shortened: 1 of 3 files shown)


MdNadimHossain (commented):
hey @vincent-gao this looks good but one question. Don't you have to update the configs in the install to match this changes as well? Else the new install site will not match these changes.

vincent-gao:
> hey @vincent-gao this looks good but one question. Don't you have to update the configs in the install to match this changes as well? Else the new install site will not match these changes.

Thanks @MdNadimHossain. Yes, that's actually the main _issue_. When the `field_group` module was introduced to Tide, no corresponding yaml config was added or updated for the form view. We can’t just update them now, because doing so would require rewriting all the related test logic~ which is clearly out of scope and would involve significant effort. At this point, we’ll likely have to live with the current setup and leave it as-is.

MdNadimHossain:
> > hey @vincent-gao this looks good but one question. Don't you have to update the configs in the install to match this changes as well? Else the new install site will not match these changes.
> 
> Thanks @MdNadimHossain. Yes, that's actually the main _issue_. When the `field_group` module was introduced to Tide, no corresponding yaml config was added or updated for the form view. We can’t just update them now, because doing so would require rewriting all the related test logic~ which is clearly out of scope and would involve significant effort. At this point, we’ll likely have to live with the current setup and leave it as-is.

What if we add the update hook functionality to the install hook as well to take care of it during install, would that work ? then no need to update any configs

MdNadimHossain (approved):
I have approved it as this fullfills the ticket requirement, but just a thought for the install hook provided

vincent-gao:
> > > hey @vincent-gao this looks good but one question. Don't you have to update the configs in the install to match this changes as well? Else the new install site will not match these changes.
> > 
> > 
> > Thanks @MdNadimHossain. Yes, that's actually the main _issue_. When the `field_group` module was introduced to Tide, no corresponding yaml config was added or updated for the form view. We can’t just update them now, because doing so would require rewriting all the related test logic~ which is clearly out of scope and would involve significant effort. At this point, we’ll likely have to live with the current setup and leave it as-is.
> 
> What if we add the update hook functionality to the install hook as well to take care of it during install, would that work ? then no need to update any configs

It would cause all the tests to fail.

PR #741, opened 2026-02-26
Feature/sd 1440 ia computed breadcrumb based on menu

MdNadimHossain:
**JIRA issue:** 
https://digital-vic.atlassian.net/browse/SD-1440
https://digital-vic.atlassian.net/browse/SD-1560

### Changed

1.  Added a new module to introduce the breadcrumb for node based on the node position in the primary and site section menu
2. A computed field for all content types that calculates the breadcrumb and shows it in the node view and it in the JSON api response
3. A tide breadcrumb builder service which follows the set requirement to build the breadcrumb trail
4. Hook to invalidate related node and menu cache so any change to dependant node or menu can be reflected in the breadcrumb trail

Test env - content-vic uat branch currently has the breadcrumb module to get it tested with section cache infront of it


modules/tide_breadcrumbs/tide_breadcrumbs.module
```diff
@@ -0,0 +1,169 @@
...
+ * @see \Drupal\tide_breadcrumbs\TideBreadcrumbBuilder
+ * @see \Drupal\tide_breadcrumbs\BreadcrumbComputedField
+ */
+function tide_breadcrumbs_node_storage_load(array $entities) {
+  if (PHP_SAPI === 'cli') {
+    return;
+  }
+
+  // Ignore the admin paths.
+  if (\Drupal::service('router.admin_context')->isAdminRoute()) {
```
MdNadimHossain, line 129: ignore triggering this hook for paths admin/content to avoid slow loading time for the view.

modules/tide_breadcrumbs/tide_breadcrumbs.module
```diff
@@ -0,0 +1,169 @@
...
+          $service->buildFullTrail($node);
+          $node->addCacheTags($service->getCacheTags($node));
+          $node->addCacheContexts($service->getCacheContexts());
+          $processed[$nid] = TRUE;
+        }
+      }
+    }
+  }
+  catch (\Exception $e) {
+    \Drupal::logger('tide_breadcrumbs')->error($e->getMessage());
```
MdNadimHossain, line 162: it is mostly to catch the error from the breadcrumb service when a site term is deleted or menu field is empty for a site and easier to debug just by looking at the dblogs

modules/tide_breadcrumbs/tide_breadcrumbs.module
```diff
@@ -0,0 +1,169 @@
...
+ * X-Drupal-Cache-Tags headers.
+ *
+ * PERFORMANCE:
+ * - Includes a CLI guard to prevent heavy breadcrumb calculation during
+ * bulk Drush operations or Cron jobs unless necessary.
+ *
+ * @see \Drupal\tide_breadcrumbs\TideBreadcrumbBuilder
+ * @see \Drupal\tide_breadcrumbs\BreadcrumbComputedField
+ */
+function tide_breadcrumbs_node_storage_load(array $entities) {
```
vincent-gao, line 123: Hi @MdNadimHossain , cacheable metadata should be attached on the computed field class (override getCacheableMetadata()), not via a global storage hook. Why are we implementing caching by adding this hook? I think, this hook is unnecessary, and I think it also has serious performance issue.

modules/tide_breadcrumbs/src/TideBreadcrumbBuilder.php
```diff
@@ -0,0 +1,598 @@
...
+      $params = $urlObj->getRouteParameters();
+      $nid = $params['node'] ?? NULL;
+
+      if ($nid) {
+        /** @var \Drupal\Core\Language\LanguageManagerInterface $language_manager */
+        $lang_code = \Drupal::languageManager()->getCurrentLanguage()->getId();
+
+        // Use direct db query to avoid triggering hook_node_storage_load().
+        // This prevents the infinite loop.
+        $query = $this->database->select('node_field_data', 'nfd');
```
vincent-gao, line 451: by removing `tide_breadcrumbs_node_storage_load` ,we no longer need this, just `Node::load($nid)->getTitle()` I guess.

vincent-gao (changes requested):
Hi @MdNadimHossain , thanks for the effort.
There are a few issues here that I think we should fix.

modules/tide_breadcrumbs/tide_breadcrumbs.module
MdNadimHossain: Hey @vincent-gao, I have explored several iterations for this change. I found that adding the cache metadata at the computed field level did not reliably invalidate the cache at the node level; the tags were not consistently bubbling up to the top-level response. Moving the logic to hook_node_storage_load ensures that cache metadata is correctly attached to the node entity itself, which is critical for both the frontend render and the JSON:API path.

To mitigate performance concerns, I have implemented several 'gatekeeper' checks: the logic is bypassed during CLI/migration operations and ignored on administrative paths (e.g., /admin/content). This ensures the overhead only occurs during legitimate node requests where the breadcrumb data is actually required.
Regarding the performance, we have covered the performance testing based on this change. Here is the ticket - https://digital-vic.atlassian.net/browse/SD-1722

modules/tide_breadcrumbs/src/TideBreadcrumbBuilder.php
MdNadimHossain: we need tide_breadcrumbs_node_storage_load to work the cahce invalidation properly in this scenario

modules/tide_breadcrumbs/tide_breadcrumbs.module
vincent-gao: Hi @MdNadimHossain , thanks for the explanation. You're right. it looks like cache tags from computed fields aren't bubbling up to the entity render array. and it seems to be fixed it in drupal 11.4 https://www.drupal.org/node/3570884. 
If that is the case, then this works for me, ✅

vincent-gao (approved):
LGTM

PR #747, opened 2026-03-04
[SD-1504] set the default metatag value for media

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-1504

### Changes
Adds a default metatag value for Media bundle.


anthony-malkoun (approved):
Nothing controversial here. Approved.

PR #752, opened 2026-03-18
[SD-1545] add node column with links

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-1545

### Change
add an extra column that displays a hyperlinked node path.


anthony-malkoun (approved):
👍

PR #753, opened 2026-03-18
[SD-992] Add "Secure file approver" role

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-992

### Change
Create a role that includes both `approver` and `secure_file_user` permissions.


anthony-malkoun (approved):
👍

PR #765, opened 2026-04-13
[SD-1598] Updated AuthorByRoleFilter to improve the slow query performance.

MdNadimHossain:
**JIRA issue:** https://digital-vic.atlassian.net/browse/SD-1706

### ISSUE
Through New Relic analysis, two primary bottlenecks were identified:

N+1 Breadcrumb Computation (56% of load): The tide_breadcrumbs logic was triggering recursive tree searches for every node in the administration list, even though breadcrumbs are not displayed there. (Fixed in the tide_breadcrumb branch)

User Entity Bloat in Filters (6.4s delay): The custom "Author by Role" filter was performing a loadMultiple() on all users matching specific roles. This triggered heavy hooks (e.g., user_expire) and massive memory consumption just to populate a dropdown list.
Screenshot 2026-04-13 at 1 07 13 pm

### Changed

1. Replaced the Entity API loadMultiple() calls with a direct db_select query. This avoids the overhead of initialising hundreds of User objects and triggering hook_user_load for every user in the system.

2. Anonymous User Handling: Switched from a full User::getAnonymousUser() entity load to a static translated string for the "Anonymous" label.

### DIFF
--- a/src/Plugin/views/filter/AuthorByRoleFilter.php
+++ b/src/Plugin/views/filter/AuthorByRoleFilter.php
@@ -121,25 +121,29 @@ class AuthorByRoleFilter extends InOperator {
     }
 
     $this->valueOptions = [];
-    // Anonymous user will display always.
-    $this->valueOptions[User::getAnonymousUser()->id()] = User::getAnonymousUser()->getDisplayName();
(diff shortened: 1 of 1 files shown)


anthony-malkoun (approved):
Yep, this seems sensible to me.

PR #771, opened 2026-05-03
[SD-1485] site section custom validation

MdNadimHossain:
**JIRA issue:** https://digital-vic.atlassian.net/browse/SD-1485

### Changed

1.  Added custom validation for field_node_site
2. The validation applies for all users except for administrator
3. Added settings to turn this on and off easily per site


vincent-gao:
@MdNadimHossain
Thank you for putting this together! But we have to cover the multiple-site scenario, which means a pretty big shift in how we build this, and we also need some kind of dynamic selection behavior.

Specifically, we need to implement the JS changes first (like a small FE), followed by the backend logic, and ensure everything is attached to the `site restriction widget`. Because this requires a major rewrite, it’s probably best to close this loop for now and open a brand new PR. Hope you don't mind, and don't hate me! 😂

PR #773, opened 2026-05-07
[SD-1869] migrate to ddev

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-1869


MdNadimHossain (changes requested):
probably we should add `ddev build` in the readme as well
and what thing I had to fix is and may be we can add this in troubleshoot doc - 
if this error happens - 
```
Failed to start tide-core: container(s) failed to become healthy before their configured timeout or in 120 seconds.
This might be a problem with the healthcheck and not a functional problem.
The error was 'ddev-tide-core-opensearch container exited.

Troubleshoot this with these commands:

  - ddev logs -s opensearch
  - docker logs ddev-tide-core-opensearch
  - docker inspect --format "{{ json .State.Health }}" ddev-tide-core-opensearch | docker run -i --rm ddev/ddev-utilities jq -r
```
look into the logs and it is because of  - 
```
ERROR: [1] bootstrap checks failed
[1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]
```
and this fixed it for me - 
`docker run --privileged --rm --pid=host alpine sysctl -w vm.max_map_count=262144`

vincent-gao:
> probably we should add `ddev build` in the readme as well and what thing I had to fix is and may be we can add this in troubleshoot doc - if this error happens -
> 
> ```
> Failed to start tide-core: container(s) failed to become healthy before their configured timeout or in 120 seconds.
> This might be a problem with the healthcheck and not a functional problem.
> The error was 'ddev-tide-core-opensearch container exited.
> 
> Troubleshoot this with these commands:
> 
>   - ddev logs -s opensearch
>   - docker logs ddev-tide-core-opensearch
>   - docker inspect --format "{{ json .State.Health }}" ddev-tide-core-opensearch | docker run -i --rm ddev/ddev-utilities jq -r
> ```
> 
> look into the logs and it is because of -
> 
> ```
> ERROR: [1] bootstrap checks failed
> [1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]
> ```
> 
> and this fixed it for me - `docker run --privileged --rm --pid=host alpine sysctl -w vm.max_map_count=262144`

Hi @MdNadimHossain I have updated the readme.md

MdNadimHossain (approved):

PR #781, opened 2026-06-02
[SD-1485] Refined site section validation

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-1485

### Changes
Adds coupling between `field_node_primary_site` and `field_node_site` on node forms. JS handles the UI part, PHP handles backend validation only. Implemented inside the existing `tide_site_restriction_field_widget`.

### Images

CleanShot 2026-06-02 at 15 28 46

https://github.com/user-attachments/assets/5507be2b-004e-4b83-a377-a40c4764929c

#### Refactor
- https://github.com/dpc-sdp/tide_core/pull/771 Abandoning this PR because the original approach didn't fully meet the requirements. Opening a fresh PR with a revised implementation


anthony-malkoun:
Given how this behaves, should we just hide the primary site field?

vincent-gao:
> Given how this behaves, should we just hide the primary site field?

Hi @anthony-malkoun This ticket's SD/AC doesn't cover hiding `field_node_primary_site`. 

what do you mean by 

> hide the primary site field?

Are you saying hide it and presave the value? 
Even if we went the "hide the field + populate it via a presave hook" route, we would still need to first define what the "primary site" actually is.  Our and QA test cases also depend on a primary site being set. I think. 

We might need some time to look into the feasibility.

anthony-malkoun (approved):
Thanks for taking me through this @vincent-gao. Approved.

PR #785, opened 2026-06-12
[SD-1801] Fixes an incompatibility issue: tide_api include resolver broken by jsonapi_resources >= 1.6

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-1801

### Issues
1. jsonapi_resources 1.6 added a decorator on the jsonapi.include_resolver service. Because tide_api.include_resolver used decoration_priority: 5, the jsonapi_resources decorator ended up as the outermost layer, so requests no longer entered TideApiIncludeResolver::resolve().

This bypassed Tide's lenient include-path handling: include paths that don't exist on every bundle (e.g. a frontend fallback like field_site_footer_logos.field_feature_image) now hit Drupal core's strict validation and return HTTP 400 instead of being silently ignored. This breaks previously-working JSON:API requests from Ripple.

2. [ResourceObjectNormalizer should declare a decoration_priority to stay innermost (silently disables jsonapi_extras enhancers)](https://git.drupalcode.org/project/jsonapi_resources/-/work_items/3599418)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -466,6 +466,9 @@
             },
             "drupal/views_bulk_operations": {
                 "getSubscribedEvents() fails on `drush updb` due to autoloader negative-cache (missing class_exists() guard) - https://www.drupal.org/project/views_bulk_operations/issues/3589629#comment-16583537": "https://www.drupal.org/files/issues/2026-05-12/views_bulk_operations-class-exists-guard-3589629-2.patch"
+            },
+            "drupal/jsonapi_resources": {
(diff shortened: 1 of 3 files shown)


lambry:
Hey @vincent-gao, I assume this is all good. 

It's just that pesky webforms test that's failing on all PRs, unfortunately that means most of the module specific tests aren't being run at all. It would be good we could get all tests running again, I might have a look at that webform issue as well.

vincent-gao:
> Hey @vincent-gao, I assume this is all good.
> 
> It's just that pesky webforms test that's failing on all PRs, unfortunately that means most of the module specific tests aren't being run at all. It would be good we could get all tests running again, I might have a look at that webform issue as well.

yeah, I am looking at the webform issue.

lambry (approved):

PR #790, opened 2026-06-22
[SD-1724] Hide section navigation title field on node form displays

vincent-gao:
### Jira
https://digital-vic.atlassian.net/browse/SD-1724

### Change
Hide section navigation title field on node form displays


anthony-malkoun (approved):
👍

